PR #3, opened 2021-07-05
Several small amendments to initial overlay spec version

MikeRalphson:
Signed-off-by: Mike Ralphson <mike.ralphson@gmail.com>

I've also put up a respec-rendered version of the spec [here](https://mermade.github.io/static/overlays_v1.0.0.html) just so we can see what needs changing in the generation script (e.g. references to OAS etc).

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -177,7 +177,7 @@ updates:
 
 #### Wildcard Overlays Examples
 
-One significant advantage of using the JMESPath syntax is that it allows referencing multiple nodes in the target document.  This would allows a single update object to be applied to multiple target objects using wildcards.
+One significant advantage of using the JMESPath syntax is that it allows referencing multiple nodes in the target document.  This would allow a single update object to be applied to multiple target objects using wildcards.
 
 ```yaml
 overlay: 1.0.0
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -274,4 +274,4 @@ The extensions may or may not be supported by the available tooling, but those m
 
 ## <a name="revisionHistory"></a>Appendix A: Revision History
 
-1.0       | TBD | First release of the Overlay Specification
+1.0.0      | TBD | First release of the Overlay Specification
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -32,7 +32,7 @@ TBD
 ## Definitions
 
 ##### <a name="overlayDocument"></a>Overlay Document
-An overlay document contains an ordered list of [Update Objects](#overlayUpdates) that are to be applied to the target document. Each [Update Object](#updateObject) has a `target` property and a modifier type (add, merge, remove).  The `target` property is a [JMESPath](http://jmespath.org/specification.html) query that identifies what part of the target document is to be updated and the modifier determines the change.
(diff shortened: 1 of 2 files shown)


versions/1.0.0.md
```diff
@@ -249,25 +251,27 @@ updates:
...
+<a name="overlayExtensions"></a>^x- | Any | Allows extensions to the Overlay Specification. The field name MUST begin with `x-`, for example, `x-internal-id`. Field names beginning `x-oai-` and `x-oas-` are reserved for uses defined by the [OpenAPI Initiative](https://www.openapis.org/). The value MAY be `null`, a primitive, an array or an object.
 
 The extensions may or may not be supported by the available tooling, but those may be extended as well to add requested support (if tools are internal or open-sourced).
 
 
 ## <a name="revisionHistory"></a>Appendix A: Revision History
 
-1.0       | TBD | First release of the Overlay Specification
\ No newline at end of file
+1.0       | TBD | First release of the Overlay Specification
```
webron, line 277:
```suggestion
1.0.0      | TBD | First release of the Overlay Specification
```
Might as well...

versions/1.0.0.md
```diff
@@ -168,14 +170,14 @@ updates:
...
-          x-safe: false
+    post:
+      description: This is an updated description of a child object
+      x-safe: false
 ```
 
 #### Wildcard Overlays Examples
 
-One significant advantage of using the JMESPath syntax that it allows referencing multiple nodes in the target document.  This would allow a single update object to be applied to multiple target objects using wildcards.
+One significant advantage of using the JMESPath syntax is that it allows referencing multiple nodes in the target document.  This would allows a single update object to be applied to multiple target objects using wildcards.
```
webron, line 180: `This allows` or `This would allow`?

versions/1.0.0.md
```diff
@@ -249,25 +251,29 @@ updates:
...
+While the Overlay Specification tries to accommodate most use cases, additional data can be added to extend the specification at certain points.
 
-The extensions properties are implemented as patterned fields that are always prefixed by `"x-"`.
+The extension properties are implemented as patterned fields that are always prefixed by `"x-"`.
 
 Field Pattern | Type | Description
 ---|:---:|---
-<a name="infoExtensions"></a>^x- | Any | Allows extensions to the OpenAPI Schema. The field name MUST begin with `x-`, for example, `x-internal-id`. Field names beginning `x-oai-` and `x-oas-` are reserved for uses defined by the [OpenAPI Initiative](https://www.openapis.org/). The value can be `null`, a primitive, an array or an object.
+<a name="overlayExtensions"></a>^x- | Any | Allows extensions to the Overlay Specification. The field name MUST begin with `x-`, for example, `x-internal-id`. Field names beginning `x-oai-` and `x-oas-` are reserved for uses defined by the [OpenAPI Initiative](https://www.openapis.org/). The value MAY be `null`, a primitive, an array or an object.
```
webron, line 270: This actually means there's a good chance there's an error in the OAS anchor for `^x-`.

webron (commented):
Changes look good, and there's no need to wait for the weekly call to merge them (it's not the core spec nor a published one).

Added a couple of comments.

The only thing I'm unsure about is the extensions - that's something we have not discussed whether we want to include in the Overlay spec, which is why I kept the content of the extension, but triple starred it (***) - it wasn't meant to be bold.

versions/1.0.0.md
MikeRalphson: Oh, I see what you mean - will check.

versions/1.0.0.md
MikeRalphson:
```suggestion
One significant advantage of using the JMESPath syntax is that it allows referencing multiple nodes in the target document.  This would allow a single update object to be applied to multiple target objects using wildcards.
```

darrelmiller (approved):

PR #17, opened 2022-03-10
Example showing setting default response

darrelmiller:


yashwason:
I dream about this feature. Would reduce a lot of copy-paste. Would also solve https://github.com/OAI/OpenAPI-Specification/issues/521

darrelmiller:
@jlk I don't usually sign my commits.  The license in the repo is authoritative when it comes to IPR.

@lornajane Do you think this would be a useful addition to Overlays?

lornajane:
Without a pull request description, this took me a while to understand. What we're saying is that we should add these fields, conditional on them not being there already? So don't update, just add if you can. I can see some use cases, but I'm not sure "default" is the right naming for this behaviour.

In an attempt to be constructive with naming, how about `add`, with wording that makes it clear it will fail if the thing is there already? Or `fallback` - something to put in if a gap is found?

lornajane:
See also #32 where we have commandeered the add syntax to mean something else. I'm still not sure what this operation should be called.

ralfhandl:
Needs to be retargeted after setting up the 2.0 dev branch

lornajane:
I don't think this pull request is useful as it is - but the idea is a very good one and merits more discussion. Let's turn it back into an issue, and work on how we could support this use case.

PR #18, opened 2022-03-10
Changed the query language from JMESPath to JSONPath

darrelmiller:
Updated the proposal to use JSONPath based on the current draft https://www.ietf.org/id/draft-ietf-jsonpath-base-04.html

JMESPath was originally selected because it had a specification and an extensive test suite.  However, experimentation has shown that JMESPath is optimized to transforming the input JSON document into an output JSON document.  The goal is to select and project.  However, for Overlays, the objective is to select a set of elements in the target document.  There is no need for the ability to project new JSON objects.

Since the original draft of the Overlay specification was created, there has been an effort in the IETF to standardize the JSONPath specification.  The original JSONPath definition relied on scripting support from the underlying implementation.  This would not be suitable for the Overlay specification.  The new IETF version is taking efforts to make the JSONPath use a fully specified query expression that is not implementation dependent.

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -32,7 +32,7 @@ TBD
 ## Definitions
 
 ##### <a name="overlayDocument"></a>Overlay Document
-An overlay document contains an ordered list of [Action Objects](#overlayActions) that are to be applied to the target document. Each [Action Object](#actionObject) has a `target` property and a modifier type (`update` or `remove`).  The `target` property is a [JMESPath](https://jmespath.org/specification.html) query that identifies what part of the target document is to be updated and the modifier determines the change.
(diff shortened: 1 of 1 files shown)


ponelat:
Link to the general page of JSONPath so I can remember: https://tools.ietf.org/wg/jsonpath/

kevinswiber (approved):
LGTM

ponelat:
@webron can you approve/merge?

lornajane:
I'm looking at implementing something to do overlays; can I assume this will be merged? I think it should be, if that counts at all!

MikeRalphson (approved):
LGTM

gregsdennis:
Uh... current draft is https://datatracker.ietf.org/doc/html/draft-ietf-jsonpath-base

PR #28, opened 2023-10-10
Fixed a leftover JMESPath instance, should be JSONPath

lornajane:
We recently switched from JMESPath to JSONPath in Overlays, but I found a stray JMESPath left in the wording. This pull request updates it.


MikeRalphson (approved):
LGTM- Thank-you

Issue #30, opened 2024-02-08
How to replace the value list of an array?


hkosova:
I want to replace the value list of an array with a new list of values. However, the two overlay implementations I've tried treat the following example as "append to an array" rather than "replace an array".

I'm not sure if the implementation behavior is correct or a bug, so wanted to confirm the intended behavior with the spec authors.

<br/>

OpenAPI document:
```yaml
openapi: 1.0.0
info:
  title: Array replacement test
  version: 1.0.0
paths: {}
x-foo:
  x-bar:  # I want to replace the values of this array
    - 1
    - 2
```

Overlay:
```yaml
overlay: 1.0.0
info:
  title: Array replacement test
  version: 0.0.1
extends: array-test.yaml

actions:
  - target: $['x-foo']
    update:
      x-bar:
        - 3
        - 4
```

<br/>

The [Action Object](https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md#action-object) section says:

> When the Overlay document is applied, the properties in the merge object replace properties in the target object with the same name

I understand this as `x-foo` being the "merge object", `x-bar` being a "property in the merge object", so the `x-bar` value from the overlay should replace (and not be appended to) the `x-bar` value in the document.

<br/>

The workaround I've found was to remove the array then re-add it with the new values. This seems a bit verbose, but if that's the intended way to replace arrays then I'm OK with it.

```yaml
actions:
  - target: $['x-foo']['x-bar']
    remove: true

  - target: $['x-foo']
    update:
      x-bar:
        - 3
        - 4
```

lornajane:
This is is a great question, and I think we don't have this use case covered. The only options are to `remove` or to` update` where `update` is described like this:

> An object with the properties and values to be merged with the object(s) referenced by the `target`. This property has no impact if `remove` property is `true`.

I'm not sure if we should add a `replace` option to cover this case. Since the spec specifically states that the list of actions is ordered and must be performed sequentially, the approach you describe of deleting and then updating is the best way to do it now (whether we add replace or not).

handrews:
@lornajane apologies if this has already been discussed as this is the first time I'm looking at Overlays, but have you looked at the [JSON Patch (RFC 6902)](https://www.rfc-editor.org/rfc/rfc6902.html) operation set?

lornajane:
[...]

PR #32, opened 2024-03-30
Switch from add to replace/update properties

lornajane:
I'm proposing this change as a response to the ongoing discussion in #30 . Using "add" can be confusing and leaves us without being able to support some use cases, such as setting an array to a particular set of elements. Looking around for similar-ish approaches, we found that JSONPath uses "update" for our existing "add" operation, and "replace" to achieve another sort of adding something. I've therefore proposed a specification update to adopt the new terminology.

It's a big change and while sooner is of course the best time to be making these changes, I'd appreciate as many eyes and as much feedback as possible. Comments and questions welcome.

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -197,7 +197,7 @@ actions:
 
 #### Array Modification Examples
 
-Array elements MAY be deleted using the `remove` property.  Use of array indexes to remove array items should be avoided where possible as indexes will change when items are removed.
+Array elements MAY be appended to existing, empty or non-existent array fields using the `update` property.
 
 ```yaml
 overlay: 1.0.0
@@ -210,6 +210,20 @@ actions:
     name: newParam
     in: query
 ```
+
+Arrays MAY be set to a set list of entries using the `replace` property. In this operation, the array will contain only the specified element(s), regardless of what the previous value of the target was.
+
+```yaml
+overlay: 1.0.0
+info:
+  title: Replace a whole array
+  version: 1.0.0
+ - target: '$.servers'
+    replace:
+      - description: Test server
+        url: https://testing.example.com/
+
+Array elements MAY be deleted using the `remove` property.  Use of array indexes to remove array items should be avoided where possible as indexes will change when items are removed.
 
 ```yaml
 overlay: 1.0.0
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -32,7 +32,7 @@ TBD
 ## Definitions
 
 ##### <a name="overlayDocument"></a>Overlay Document
-An overlay document contains an ordered list of [Action Objects](#overlayActions) that are to be applied to the target document. Each [Action Object](#actionObject) has a `target` property and a modifier type (`update` or `remove`).  The `target` property is a [JSONPath](https://datatracker.ietf.org/wg/jsonpath/documents/) query expression that identifies the elements of the target document to be updated and the modifier determines the change.
+An overlay document contains an ordered list of [Action Objects](#overlayActions) that are to be applied to the target document. Each [Action Object](#actionObject) has a `target` property and a modifier type (`add`, `replace`, or `remove`).  The `target` property is a [JSONPath](https://datatracker.ietf.org/wg/jsonpath/documents/) query expression that identifies the elements of the target document to be updated and the modifier determines the change.
 
 ## Specification
 
@@ -80,7 +80,7 @@ Field Name | Type | Description
 <a name="overlayVersion"></a>overlay | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the Overlay Specification that the Overlay document uses. The `overlay` field SHOULD be used by tooling to interpret the Overlay document. 
 <a name="overlayInfo"></a>info | [Info Object](#infoObject) | **REQUIRED**. Provides metadata about the Overlay. The metadata MAY be used by tooling as required.
 <a name="overlayExtends"></a> extends | `string` | URL to the target document (such as an OpenAPI document) this overlay applies to. This MUST be in the form of a URL.
-<a name="overlayActions"></a>actions | [[Action Object](#actionObject)] | **REQUIRED** An ordered list of actions to be applied to the target document. The array MUST contain at least one value.
+<a name="overlayActions"></a> actions | [[Action Object](#actionObject)] | **REQUIRED** An ordered list of actions to be applied to the target document. The array MUST contain at least one value.
 
 This object MAY be extended with [Specification Extensions](#specificationExtensions).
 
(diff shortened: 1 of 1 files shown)


philsturgeon:
Typo in lorna's message there, she meant to say:

> we found that JSON Pat**c**h

Here's a direct link to the operations we could model Overlay operations off of, either partially or entirely: https://www.rfc-editor.org/rfc/rfc6902.html#section-4

versions/1.0.0.md
```diff
@@ -111,10 +111,11 @@ This object represents one or more changes to be applied to the target document
...
+target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
+description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
+add | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. If an object property exists, it is overwritten with the new value. Array elements are appended to any existing entries already in the array. This property MUST NOT be used with the `replace` property in the same action object. This property has no impact if `remove` property is `true`.
+replace | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. The entire contents of the target object or array is replaced by the new value(s) supplied. This property MUST NOT be used with the `add` property in the same action object. This property has no impact if `remove` property is `true`.
```
philsturgeon, line 117: Is this right? It says the values will be merged into the target then it the values will be replaced in the target.

versions/1.0.0.md
```diff
@@ -111,10 +111,11 @@ This object represents one or more changes to be applied to the target document
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="actionTarget"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
-<a name="actionDescription"></a>description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="actionUpdate"></a>update | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. This property has no impact if `remove` property is `true`.
```
philsturgeon, line 116: Is it worth keeping update aroud for a minute as deprecated or is this early and experimental enough that whatever?

kevinswiber: This is always a hard decision to make.  At this point in time, I'd vote to just kill it.  We won't always have this luxury, but we should make use of it while we can.

kevinswiber:
Really appreciate having some parallels with JSON Patch!

versions/1.0.0.md
```diff
@@ -196,7 +197,7 @@ actions:
 
 #### Array Modification Examples
 
-Array elements MAY be deleted using the `remove` property.  Use of array indexes to remove array items should be avoided where possible as indexes will change when items are removed.
+Array elements MAY be appended to existing, empty or non-existent array fields using the `update` property.
```
darrelmiller, line 200:
```suggestion
Array elements MAY be appended to existing, empty or non-existent array fields using the `add` property.
```
Is this what was intended?

versions/1.0.0.md
darrelmiller: Agreed, the first sentence for replace seems to contradict the second.

versions/1.0.0.md
lornajane: We'll always have the commit history, but this does make me wonder if we should propose a 0.1.0 version and then start following a more formal release process. We're already seeing usage ....

versions/1.0.0.md
lornajane: Updated, could you check this makes more sense?

versions/1.0.0.md
lornajane: .... Alternatively, how about moving forward from our current situation by making this a 1.1.0 and still a draft?

versions/1.0.0.md
```diff
@@ -111,10 +111,11 @@ This object represents one or more changes to be applied to the target document
 
 Field Name | Type | Description
 ---|:---:|---
-<a name="actionTarget"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
```
kevinswiber, line 114: I noticed that anchors are removed.  I think they don't quite work well with GitHub's formatting, anyway, but it's worth pointing out. Not sure if this is used with other tooling.  Not a blocker.

kevinswiber (approved):
LGTM

kevinswiber:
> @kevinswiber Thanks for the approval, but do you have any thoughts on how to version this in a way that won't confuse existing users? That's what got me really stuck.

@lornajane Seeing as how Overlays has been in a pre-release phase, I'm not compelled to allow for backwards-compatibility.  Breaking backwards-compatibility has a negative impact on early users but a positive impact on future users.  I look at an example of a project that made this hard decision, Node.js.  Before their first major release, they'd break backwards compatibility all the time.  This was becoming more difficult as adoption was growing.  In retrospect, I believe it was the right move.  It led to greater stability when even _more_ users were being impacted by changes.  

I believe @ThomasRooney has a different perspective worth including in the conversation, as they're dealing with existing users.

ThomasRooney:
> I believe @ThomasRooney has a different perspective worth including in the conversation, as they're dealing with existing users.

Thanks for asking. My persective is that making a breaking change without a corresponding version change is always bad, the moment that specification has been adopted. I understand that this specification remains a "draft", but it is also:

1. Over 3 years old.
2. Already in use by a non-trivial number of organisations, given it solves an significant existing need. Whilst a few of these organisations (e.g. Speakeasy, Bump) work in public, through my work with Speakeasy I can definitely state that there are a number of organisations that have also adopted the overlay specification internally that would also be affected.

There's nothing wrong with breaking changes, but I strongly feel that if a breaking change is deemed important, it would be kinder to:

1. Leave the current document as `1.0.0`.
2. Make any breaking changes in `2.0.0`

Or, alternatively:

1. Reduce the current specification version to `0.1.0`.
2. Make any breaking changes wanted in `0.2.0`.

That way, any of the organisations that manage documents with `overlay: 1.0.0` get an easy migration path: we'd just automatically alter the version of any overlay documents we interact with from `1.0.0` to `0.1.0`. This pathway would remain viable for however many months (years?) it'll take to make a non-draft `1.0.0`.

----------

I personally think the change from `add` to `update` isn't actually a positive one anyway, because:

1. An openapi overlay is data-format-agnostic. I.e. it doesn't just apply to JSON documents, but to whatever format an openapi document is kept in. This, whilst primarily YAML, also includes [pkl](https://github.com/apple/pkl), [cue](https://cuelang.org/), as well as in-memory representations. Hence I have a bit of aversion to leaning on JSON Patch `add` nomenclature to respresent the "shallow merge + array append" that is actually happening.
2. Probably the most common thing we see overlays used for is overriding something like `description`. This is primarily by a DX/technical writing team trying to produce a good documentation site or SDK type info that's produced downstream of the OAS. Whilst this is usually done by a shallow merge of the parent object, we've also seen `target` be applied to the scalar string value. `add` feels ambiguous when applied to scalars (strings, numbers), whereas `update` does not. E.g. consider the case where a `description` node is targetted to be overridden, `add` sounds to me like it should append to the existing string, rather than override it.

For the underlying problem:
1. In my opinion, adding a `replace` action would be a nice bit of syntactic sugar to reduce ambiguity, especially for cases where an array item is being replaced.
2. Would also love to have a `move` action for something like adding a `oneOf` (or converting a `oneOf` into an `allOf`/`anyOf`) in a JSON Schema. This is a relatively common action for us as we help represent different use-cases with more constrained json schemas to give better DX for our customers APIs, but we currently risk a higher degree of brittleness to do this with `update` overrides.

handrews:
@ThomasRooney 

> An openapi overlay is data-format-agnostic. I.e. it doesn't just apply to JSON documents, but to whatever format an openapi document is kept in. This, whilst primarily YAML, also includes [pkl](https://github.com/apple/pkl), [cue](https://cuelang.org/), as well as in-memory representations.

This is a good point to support my recent proposal that OAS 4.0 "Moonwalk" should define an in-memory structure and how it can (and can't) be manipulated, so that we're less format-dependent for programmatic editing.  Which doesn't change anything for this PR given how long it will be before 4.0 is a practical concern.  But I feel like the Overlays project should impact that design.

versions/1.0.0.md
philsturgeon: Question retracted, you folks came up with a god solution on the call and I forget what it was, but I'm not standing in the way of anything 🫡

versions/1.0.0.md
philsturgeon: @lornajane smashed it, makes perfect sense, resolve me!

versions/1.0.0.md
```diff
@@ -210,6 +211,20 @@ actions:
...
+overlay: 1.0.0
+info:
+  title: Replace a whole array
+  version: 1.0.0
+ - target: '$.servers'
+    replace:
+      - description: Test server
+        url: https://testing.example.com/
+
+Array elements MAY be deleted using the `remove` property.  Use of array indexes to remove array items should be avoided where possible as indexes will change when items are removed.
```
ralfhandl, line 226: What is the recommended way of removing certain array elements?

versions/1.0.0.md
lornajane: This proposal will definitely be 2.0 in recognition of how breaking it is.

ralfhandl:
Needs to be retargeted after setting up the 2.0 dev branch

lornajane:
We will need to revisit this for a future major release, there's an open issue already so I'll close this pull request.

PR #47, opened 2024-09-01
Add an intro and other missing sections

### LINKED ISSUE #44: Write an introduction for the spec
It says "TBD" in the introduction section, we should write some words.

lornajane:
We had some "TBD" sections, I removed one (data types, I don't think we can add this after so many people have implemented, and Overlays could span multiple OpenAPI versions!?!?), wrote an intro, and used a version of the Arazzo wording, acknowledging the retroactive nature of the 1.0.0 version specifically.

Hopefully this fills the main gaps! Fixes #44

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -11,9 +11,9 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 The Overlay specification is an extension or companion to the OpenAPI specification.
 An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI description.
 
-The main purpose of Overlays is to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
+The main purpose of the Overlay Specification is to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
 Use cases include updating descriptions, adding metadata to be consumed by another tool, or removing certain elements from an API description before sharing it with partners.
-Overlays may be specific to a single OpenAPI description, or be designed to apply the same transform to any OpenAPI description.
+An Overlay may be specific to a single OpenAPI description, or be designed to apply the same transform to any OpenAPI description.
 
 ## Table of Contents
 
@@ -35,8 +35,9 @@ Overlays may be specific to a single OpenAPI description, or be designed to appl
 
 ## Definitions
 
-##### <a name="overlayDocument"></a>Overlay Document
-An overlay document contains an ordered list of [Action Objects](#overlayActions) that are to be applied to the target document. Each [Action Object](#actionObject) has a `target` property and a modifier type (`update` or `remove`).  The `target` property is a [JSONPath](https://datatracker.ietf.org/wg/jsonpath/documents/) query expression that identifies the elements of the target document to be updated and the modifier determines the change.
+##### <a name="overlayDocument"></a>Overlay
+
+An Overlay is a JSON or YAML structure containing an ordered list of [Action Objects](#overlayActions) that are to be applied to the target document. Each [Action Object](#actionObject) has a `target` property and a modifier type (`update` or `remove`).  The `target` property is a [JSONPath](https://datatracker.ietf.org/wg/jsonpath/documents/) query expression that identifies the elements of the target document to be updated and the modifier determines the change.
 
 ## Specification
 
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -11,7 +11,7 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 The Overlay specification is an extension or companion to the OpenAPI specification.
 An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI document.
 
-The main purpose of Overlays to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
+The main purpose of Overlays is to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
 Use cases include updating descriptions, adding metadata to be consumed by another tool, or removing certain elements from an API description before sharing it with partners.
 Overlays may be specific to a single OpenAPI description, or be designed to apply the same transform to any OpenAPI description.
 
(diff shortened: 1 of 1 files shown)


philsturgeon (approved):

versions/1.0.0.md
```diff
@@ -8,7 +8,12 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 
 ## Introduction
 
-TBD
+The Overlay specification is an extension or companion to the OpenAPI specification.
+An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI document.
+
+The main purpose of Overlays to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
```
ralfhandl, line 14:
```suggestion
The main purpose of Overlays is to provide a way to repeatably apply transformations to one or many OpenAPI descriptions.
```

versions/1.0.0.md
```diff
@@ -38,7 +42,10 @@ An overlay document contains an ordered list of [Action Objects](#overlayActions
 
 ### Versions
 
-TBD
+The Overlay Specification is versioned using a `major`.`minor`.`patch` versioning scheme. The `major`.`minor` portion of the version string (for example 1.0) SHALL designate the Arazzo feature set. `.patch` versions address errors in, or provide clarifications to, this document, not the feature set. The patch version SHOULD NOT be considered by tooling, making no distinction between 1.0.0 and 1.0.1 for example.
```
ralfhandl, line 45:
```suggestion
The Overlay Specification is versioned using a `major`.`minor`.`patch` versioning scheme. The `major`.`minor` portion of the version string (for example 1.0) SHALL designate the Overlays feature set. `.patch` versions address errors in, or provide clarifications to, this document, not the feature set. The patch version SHOULD NOT be considered by tooling, making no distinction between 1.0.0 and 1.0.1 for example.
```

Which raises the question whether "Overlays" needs to to through a naming approval process, and whether it will come out as "Intonaco" or "Copertura" or something similar.

versions/1.0.0.md
```diff
@@ -38,7 +42,10 @@ An overlay document contains an ordered list of [Action Objects](#overlayActions
...
+
+**Note:** Version 1.0.0 of the Overlay Specification is being released after spending some time in draft, and after being adopted by tools providers.
```
ralfhandl, line 47:
```suggestion
**Note:** Version 1.0.0 of the Overlay Specification is being released after spending some time in draft, and after being adopted by tool providers.
```
Sounds better even if there probably are providers providing multiple tools.

versions/1.0.0.md
lornajane: Let's try to avoid that.

versions/1.0.0.md
```diff
@@ -8,7 +8,12 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 
 ## Introduction
 
-TBD
+The Overlay specification is an extension or companion to the OpenAPI specification.
+An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI document.
```
kevinswiber, line 12: Like with OAS, we'll probably need a way to differentiate between Overlay Specification, an Overlay document, and the conceptual manifestation that may be a singular document or a cohesion of documents (Overlay Description?). I would recommend adopting the patterns we have with OpenAPI already if they make sense, but I'm open to other opinions on this.

versions/1.0.0.md
kevinswiber, line 45: I understand this is what Arazzo is using, but we're currently in discussions on that project regarding what version bumps really mean.  I'm not sure a `SHOULD NOT` reflects the reality.  For example, in Arazzo right now, we're discussing a change in grammar of simply moving a set of parentheses.  That small clarification might constitute a patch change in the spec, but toolmakers may have to consider it, as they could have misinterpreted the spec the first time around. I understand the need to assuage concerns, but I might just leave it out for now.

kevinswiber:
Thank you for doing this, @lornajane!

versions/1.0.0.md
ralfhandl: Honestly: "simply moving a set of parentheses" is a breaking/major change unless the spec allows both placements of parentheses and only calls out which placement is preferred/future-proof.

Adding a feature is a minor change.

Rephrasing text without adding or removing a feature is a patch.

versions/1.0.0.md
lornajane: I'm not sure what you're suggesting in terms of rewording, can you clarify if there's a change you had in mind?

I used "Overlay" pretty intentionally, but I now think the end should be "OpenAPI description" and I'll update.

versions/1.0.0.md
lornajane: I'd be happy to drop or simplify this section, but we'll need to clarify it some time and I thought this was pretty lightweight and close to the way we usually manage specifications. Does anyone have a counter-wording to suggest?

versions/1.0.0.md
```diff
@@ -8,7 +8,12 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 
 ## Introduction
 
-TBD
+The Overlay specification is an extension or companion to the OpenAPI specification.
+An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI description.
```
ralfhandl, line 12:
```suggestion
An Overlay document describes a set of changes to be applied or "overlaid" onto an existing OpenAPI description.
```

ralfhandl (changes requested):
+1, minor nits only

versions/1.0.0.md
ralfhandl: We don't have the `$ref` problems here that requires the distinction between a "document" (single file) and "description" (one or more files linked via `$ref`, so it would suffice to always write either "Overlay Specification" or "Overlay document" and avoid stand-alone occurrences of "Overlay" or "Overlays".

Have applied that as suggestions here.

If accepted, we need another pass over the spec to adjust places not changed by this PR.

mikekistler (dismissed):
Looks good! 👍

ralfhandl (approved):

ralfhandl (approved):

versions/1.0.0.md
```diff
@@ -8,19 +8,27 @@ This document is licensed under [The Apache License, Version 2.0](https://www.ap
 
 ## Introduction
 
-TBD
+The Overlay Specification is an extension or companion to the [[OpenAPI]] specification.
+An Overlay describes a set of changes to be applied or "overlaid" onto an existing OpenAPI description.
```
mikekistler, line 12: Do we want to restrict Overlays to only operate on OpenAPI documents? For the most part the spec could easily be applied to any JSON or YAML doc. And in several places we seem to allow base documents of other types, e.g. target is described as
> URL to the target document (such as an [[OpenAPI]] document) ...

mikekistler (commented):
This mostly looks good, but I think we need to be clear about whether Overlays can be used on JSON or YAML docs _other than_ OpenAPI descriptions.

versions/1.0.0.md
ralfhandl: The repo README.md says

> The Overlay specification defines a way of creating documents that contain information to be merged with an OpenAPI description at some later point in time, for the purpose of updating the OpenAPI description with additional information.

Given that with [RFC6902](https://datatracker.ietf.org/doc/html/rfc6902) and [RFC7386](https://datatracker.ietf.org/doc/html/rfc7386) there are already two document formats for modifying arbitrary JSON documents, I'd keep to the original scope.

@kevinswiber @lornajane?

ralfhandl: @mikekistler How about merging this PR and opening a dedicated issue for defining the scope of Overlay target documents?

PR #50, opened 2024-09-01
Clarify the update behaviour for arrays as well as objects

lornajane:
As discussed in a recent meeting [see notes](https://github.com/OAI/Overlay-Specification/discussions/45#discussioncomment-10465119), clarify that for arrays, the update adds an entry but for objects the items are merged ( as is already described)

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -113,7 +113,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="actionTarget"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
 <a name="actionDescription"></a>description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="actionUpdate"></a>update | Any | If the `target` is an object, an object with the properties and values to merge with the `target`. If the `target` is an array, an entry to append to the array. To replace an array in its entirety, apply a remove action followed up an update action. This property has no impact if `remove` property is `true`.
+<a name="actionUpdate"></a>update | Any | If the `target` of this action object is an object, this field should be an object with the properties and values to merge with the `target`. If the `target` of this action is an array, this field should be an entry to append to the array. To replace an array in its entirety, apply a remove action followed by an update action. This field has no impact if the `remove` field of this action object is `true`.
 <a name="actionRemove"></a>remove | `boolean` | A boolean value that indicates that the target object is to be removed from the the map or array it is contained in. The default value is `false`.
 
 The result of the `target` JSONPath query expression must be zero or more objects or arrays (not primitive types or `null` values). If you wish to update a primitive value such as a string, the `target` expression should select the *containing* object in the target document.
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -93,7 +93,7 @@ This object represents one or more changes to be applied to the target document 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
-| <a name="action-description"></a>description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation. |
+| <a name="action-description"></a>description | `string` | A description of the action. [[CommonMark]] syntax MAY be used for rich text representation. |
 | <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field should be an object with the properties and values to merge with the node. If the `target` selects an array, the value of this field should be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true`. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object is to be removed from the the map or array it is contained in. The default value is `false`. |
 
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -97,7 +97,11 @@ This object represents one or more changes to be applied to the target document 
 | <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field should be an object with the properties and values to merge with the node. If the `target` selects an array, the value of this field should be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true`. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object is to be removed from the the map or array it is contained in. The default value is `false`. |
 
-The result of the `target` JSONPath expression must be zero or more objects or arrays (not primitive types or `null` values). If you wish to update or remove a primitive value such as a string, the `target` expression should select the _containing_ object in the target document.
+The result of the `target` JSONPath expression must be zero or more objects or arrays (not primitive types or `null` values).
+
+To update a primitive property value such as a string, the `target` expression should select the _containing_ object in the target document and `update` should contain an object with the property and its new primitive value.
+
+Primitive-valued items of an array cannot be replaced or removed individually, only the complete array can be replaced.
 
 The properties of the update object MUST be compatible with the target object referenced by the JSONPath key. When the Overlay document is applied, the properties in the merge object replace properties in the target object with the same name and new properties are appended to the target object.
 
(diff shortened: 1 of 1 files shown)


ralfhandl (dismissed):

versions/1.0.0.md
```diff
@@ -113,7 +113,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="actionTarget"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
 <a name="actionDescription"></a>description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="actionUpdate"></a>update | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. This property has no impact if `remove` property is `true`.
+<a name="actionUpdate"></a>update | Any | If the `target` is an object, an object with the properties and values to merge with the `target`. If the `target` is an array, an entry to append to the array. To replace an array in its entirety, apply a remove action followed up an update action. This property has no impact if `remove` property is `true`.
```
kevinswiber, line 116: I think this perhaps requires more context below this table where we can create additional prose. I'm wondering if it should be reworked to give us more space for explanation.

Apologies for being pedantic below.

- `target` needs to be better defined in this context.  Is it the JSONPath query itself? Is it the document the query is using as its source?  Is it the result of the query on the document? What happens if a result contains multiple entries of mixed types?
- JSON has "members" or "name/value pairs", not "properties".
- "To replace an array in its entirety, apply a remove action followed up an update action. This property has no impact if `remove` property is true." This is a little confusing.  I think it means "followed _by_ an update action". The last sentence seemingly contradicts the first, and I think it can be rephrased with more clarity.

I would propose keeping it short here and adding more detail below where we can expand beyond the table format.

lornajane: Easiest question first, the `target` is the thing two lines up, the target of this action.

I definitely didn't nail the field/property wording, I'll give that another go.

versions/1.0.0.md
```diff
@@ -113,7 +113,7 @@ Field Name | Type | Description
 ---|:---:|---
 <a name="actionTarget"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document.
 <a name="actionDescription"></a>description | `string` | A description of the action. [CommonMark syntax](https://spec.commonmark.org/) MAY be used for rich text representation.
-<a name="actionUpdate"></a>update | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. This property has no impact if `remove` property is `true`.
+<a name="actionUpdate"></a>update | Any | If the `target` of this action object is an object, this field should be an object with the properties and values to merge with the `target`. If the `target` of this action is an array, this field should be an entry to append to the array. To replace an array in its entirety, apply a remove action followed by an update action. This field has no impact if the `remove` field of this action object is `true`.
```
ralfhandl, line 116: > To replace an array in its entirety, apply a remove action followed by an update action.

This can work if the target is an OAS-defined fixed field or a JSON-Schema-defined keyword and the tool knows whether the target is array- or object-typed.

Or if the tool "remembers" the before-type of target that it deleted. In which case one can't replace an array with a single object, only with a one-item array containing that object.

Is that a real problem?

ralfhandl (dismissed):

mikekistler (approved):
Looks good! 👍

ralfhandl (approved):

versions/1.0.0.md
```diff
@@ -92,12 +92,12 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field should be an object with the properties and values to merge with the node. If the `target` selects an array, the value of this field should be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true`. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object is to be removed from the the map or array it is contained in. The default value is `false`. |
 
-The result of the `target` JSONPath query expression must be zero or more objects or arrays (not primitive types or `null` values). If you wish to update a primitive value such as a string, the `target` expression should select the _containing_ object in the target document.
+The result of the `target` JSONPath expression must be zero or more objects or arrays (not primitive types or `null` values). If you wish to update or remove a primitive value such as a string, the `target` expression should select the _containing_ object in the target document.
```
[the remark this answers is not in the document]
ralfhandl: You replace the array with a new one that does not contain the item to remove. Which means you have to know the full array content.

I definitely want to drop this whole paragraph and allow `target` expressions to select any kind of node, including primitive values, but that would be a significant change since the restriction to objects and arrays has been in since the spec document was created more than three years ago.

Opened
* #57 
to track and discuss this.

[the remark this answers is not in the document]
mikekistler: Then can we just go back to the original wording of that last sentence:
> If you wish to update a primitive value such as a string, the `target` expression should select the _containing_ object in the target document.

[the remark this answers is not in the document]
ralfhandl: The second example in [Array Modification Examples](https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md#array-modification-examples) shows how to remove an object from an array:
~~~yaml
overlay: 1.0.0
info:
  title: Remove a array element
  version: 1.0.0
actions:
  - target: $.paths.*.get.parameters[?@.name == 'dummy']
    remove: true
~~~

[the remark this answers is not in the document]
mikekistler: Got it. You can remove an array element when it is an object using JSON Path filtering.

So do we go back to the original wording for that last sentence? By original wording I mean:
> If you wish to update a primitive value such as a string, the `target` expression should select the _containing_ object in the target document.

We _could_ update this to be even more explicit:
> If you wish to update a primitive value such as a string, the `target` expression should select the _containing_ object in the target document and the action should contain the property with the value to be updated.

ralfhandl, line 100:
```suggestion
The result of the `target` JSONPath expression must be zero or more objects or arrays (not primitive types or `null` values).

To update a primitive property value such as a string, the `target` expression should select the _containing_ object in the target document and `update` should contain an object with the property and its new primitive value.

Similarly for arrays of primitive values: to replace or remove a primitive array item, the `target` expression should select the  object containing the array-valued property and `update` should contain an object with the property and its new array value.
```

versions/1.0.0.md
```diff
@@ -92,12 +92,12 @@ This object represents one or more changes to be applied to the target document
 
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
-| <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath query expression referencing the target objects in the target document. |
+| <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
 | <a name="action-description"></a>description | `string` | A description of the action. [[CommonMark]] syntax MAY be used for rich text representation. |
-| <a name="action-update"></a>update | Any | An object with the properties and values to be merged with the object(s) referenced by the `target`. This property has no impact if `remove` property is `true`. |
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field should be an object with the properties and values to merge with the node. If the `target` selects an array, the value of this field should be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true`. |
```
ralfhandl, line 97: Is the "merge" shallow or deep? That is, if `target` selects an object and `update` is an object with an object-valued property, is that property's object value _replaced_, or is it also (recursively) _merged_?

versions/1.0.0.md
mikekistler: I think the operation you describe in the last paragraph is simply a replace of an array. Characterizing this as replacing or removing a primitive array item seems misleading to me.

ralfhandl: Ok, how about the modified proposal?

mikekistler: That looks good! Thank you. Pls commit that and then I'll give it one last pass.

versions/1.0.0.md
mikekistler: On line 106 below it says:
> When the Overlay document is applied, the properties in the merge object replace properties in the target object with the same name and new properties are appended to the target object.

Based on that wording, I think the "merge" is actually a shallow merge.

mikekistler (approved):
Looks good! 👍

ralfhandl (approved):

PR #55, opened 2024-09-09
Add respec workflow

ralfhandl:
Adds workflow to regenerate HTML versions of the Overlay specification

To Do for follow-up PR:
* Create PRs in OAI/OpenAPI-Specification repo once Overlay 1.0 is released

Currently PRs will be created in
* https://github.com/ralfhandl/OpenAPI-Specification/pulls

and will deploy to 
* https://ralfhandl.github.io/OpenAPI-Specification/overlay/v1.0.0.html

which allows us to proof-read the respec-formatted HTML.

### DIFF
--- a/.github/workflows/respec.yaml
+++ b/.github/workflows/respec.yaml
@@ -54,6 +54,7 @@ jobs:
         labels: Housekeeping
         # reviewers: OAI/tsc
         author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+        committer: GitHub <noreply@github.com>
         title: Update ReSpec-rendered specification versions for Overlay
         commit-message: Update ReSpec-rendered specification versions
         signoff: false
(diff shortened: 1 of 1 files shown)


ralfhandl:
> TODOs are fixed before merge

TODOs that can be fixed before the release have been fixed.

.github/workflows/respec.yaml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        node-version: '20.x'
+    
+    - name: Install dependencies
+      run: npm ci
+
+    - uses: actions/checkout@v4 # checkout gh-pages branch
+      with:
+        token: ${{ secrets.OAS_REPO_TOKEN }}
+        repository: ralfhandl/OpenAPI-Specification  # TODO: change to OAI/...
```
mikekistler, line 38: Based on your comment, I assume that means this TODO must be fixed after merge. I'll take your word on that ...

mikekistler (approved):
Looks good! 👍

.github/workflows/respec.yaml
ralfhandl: Correct, otherwise we'd open PRs in https://github.com/OAI/OpenAPI-Specification/pulls for publishing the Overlay specification on https://spec.openapis.org/, which we probably don't want before we release v1.0.

PR #62, opened 2024-09-25
Minor nits

ralfhandl:


versions/1.0.0.md
```diff
@@ -44,7 +44,7 @@ In order to preserve the ability to round-trip between YAML and JSON formats, [[
 
 ### Document Structure
 
-It is RECOMMENDED that the root Overlay document be named: `overlay.json` or `overlay.yaml`.
+It is RECOMMENDED that an Overlay document be named: `overlay.json` or `overlay.yaml`.
```
ralfhandl, line 47: Overlay documents can't reference other documents, so there is no "root".

Also there may be multiple overlays to be applied to the same target OpenAPI Description: should all of them be named `overlay.yaml`, thus requiring a folder strucure?

Should we rather remove this whole section?

lornajane: Yes, let's remove. I don't see value in dictating file names.

mikekistler (approved):
Looks good! 👍

PR #72, opened 2024-10-08
feat: add JSON Schema 2020-12 schemas

### LINKED ISSUE #71: JSON Schema 2020-12 schema
I've updated the v1.0 schema

One clarification, the `allOf` for `update` and `remove` scenario
https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/versions/1.0.0.md?plain=1#L122

> This field has no impact if the `remove` field of this action object is `true`.

Is this something that needs to be enforced with validation?

jeremyfiel:
closes #71

### DIFF
--- a/schemas/v1.0/readme.md
+++ b/schemas/v1.0/readme.md
@@ -1,0 +1,36 @@
+# OpenAPI Overlay 1.0.x JSON Schema
+
+Here you can find the JSON Schema for validating OpenAPI Overlay description of versions 1.0.x.
+
+As a reminder, the JSON Schema is not the source of truth for the Specification.
+In cases of conflicts between the Specification itself and the JSON Schema, the
+Specification wins. Also, some Specification constraints cannot be represented
+with the JSON Schema so it's highly recommended to employ other methods to
+ensure compliance.
+
+The iteration version of the JSON Schema can be found in the `$id` field.
+For example, the value of `$id: https://spec.openapis.org/overlay/1.0/schema/2024-10-17` means this iteration was created on October 17, 2024.
+
+## Contributing
+
+To submit improvements to the schema, modify the `schema.yaml` and add test cases for your changes.
+
+The TSC will then:
+- Run tests on the updated schema
+- Update the iteration version
+- Publish the new version
+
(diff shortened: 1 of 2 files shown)


schemas/v1/schema.json
ralfhandl, line 1: I suggest following the folder name convention in https://github.com/OAI/OpenAPI-Specification/tree/main/schemas and rename this folder to `v1.0`

jeremyfiel:
updated the schema to include a validation check for `action-object`. 

> If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node.

---

Because an `array` update can be of `type: any`, I did not include an additional validation check for that scenario outlined as a `MUST`. 
> If the `target` selects an array, the value of this field MUST be an entry to append to the array.

schemas/v1.0/schema.json
ralfhandl, line 1: I'd remove this file as we will most likely have a workflow for generating it on the fly for publication, similar to
* https://github.com/OAI/OpenAPI-Specification/pull/4139

ralfhandl (commented):
In addition to this we'll also need (not necessarily in this PR)
- [ ] tests for the schema, similar to https://github.com/OAI/OpenAPI-Specification/tree/main/tests/v3.1
- [ ] A `schemas/README.MD` file similar to https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.1/README.md, just one folder down directly in `schemas`

jeremyfiel:
related #86

schemas/v1.0/readme.md
```diff
@@ -0,0 +1,36 @@
...
+
+```bash
+npm install
+npm test
+```
+
+You can also validate a document individually.
+
+```bash
+node scripts/validate.mjs path/to/document/to/validate.yaml
```
ralfhandl, line 35: Oops, forgot to copy this 🙄

ralfhandl: Added: thanks for reminding me!

PR #76, opened 2024-10-12
Updated use of URL to URI

darrelmiller:
This PR attempts to clean up some of the language that was copied from OAS that we have since realized was ambiguous.  It also changes the specification to use URI references instead of URLs for the field extends.

The reason for this change is that we want to give the option for tooling to choose whether to treat the URI reference as an identifier or a locator.  With the previous wording a tooling author may interpret the specification as saying that the tooling is required to dereference the extends URI reference to acquire the target document.  That wouldn't be good, for security reasons and it could lead to users to assume that they can pass an overlay with the extends keyword to tooling that expect to receive an OpenAPI document, because it appears that the Overlay is self contained. (This is similar to how Kubernetes Kustomize overlays can be used https://kubectl.docs.kubernetes.io/guides/config_management/introduction/)

This aligns Overlay's use of URI references with the OAS's use of URI references.

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -73,7 +73,7 @@ The list of actions MUST be applied in sequential order to ensure a consistent o
 
 The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] document. Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI document(s).
 
-In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document identified by an absolute URI reference.
+In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document, identified by an absolute URI reference.
 
 ```yaml
 overlay: 1.0.0
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -73,7 +73,7 @@ The list of actions MUST be applied in sequential order to ensure a consistent o
 
 The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] document. Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI document(s).
 
-In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document, identified by an absolute URI reference.
+In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document, identified by an absolute URI.
 
 ```yaml
 overlay: 1.0.0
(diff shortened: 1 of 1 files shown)


versions/1.0.0.md
```diff
@@ -74,7 +73,7 @@ The list of actions MUST be applied in sequential order to ensure a consistent o
 
 The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] document. Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI document(s).
 
-In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document using an absolute URL.
+In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document, identified by an absolute URI reference.
```
handrews, line 76: You can just say "absolute URI" here as an absolute URI is by definition not a relative reference.  "absolute URI reference" is still technically correct (since "URI reference" encompasses all other URI terms) but sounds weird to me.

I don't feel strongly about this so feel free to leave it.

```suggestion
In the following example the `extends` property specifies that the overlay is designed to update the OpenAPI Tic Tac Toe example document, identified by an absolute URI.
```

handrews (approved):
One minor nit but otherwise this looks great!

ralfhandl (approved):

ralfhandl:
Committed @handrews' suggestion

versions/1.0.0.md
```diff
@@ -46,10 +46,9 @@ In order to preserve the ability to round-trip between YAML and JSON formats, [[
 - Tags MUST be limited to those allowed by the [JSON Schema ruleset](https://yaml.org/spec/1.2/spec.html#id2803231).
 - Keys used in YAML maps MUST be limited to a scalar string, as defined by the [YAML Failsafe schema ruleset](https://yaml.org/spec/1.2/spec.html#id2802346).
 
-### Relative References in URLs
+### Relative References in URIs
 
-Unless specified otherwise, all properties that are URLs MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
-Unless specified otherwise, relative references are resolved using the URL of the referring document.
+Unless specified otherwise, all fields that are URI references MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
```
mikekistler, line 51: Do we need to say what the Base URI is here?  I think we do in the OAS.

mikekistler (approved):
The changes here seem fine, though I'm not really sure how they address this concern:
> it could lead to users to assume that they can pass an overlay with the extends keyword to tooling that expect to receive an OpenAPI document, because it appears that the Overlay is self contained. 

Also I wonder if we need to define the base URI for relative references as we do in OAS.

versions/1.0.0.md
handrews: @mikekistler no, because RFCe3986 takes care of it, and the Overlay Specification does not define an internal base URI-setting mechanism under RFC3986 §5.1.1.  Most of the weird problems regarding base URIs in the OpenAPI Specification are because someone tried to re-write things already covered by RFC3986 and messed it up (e.g. forgetting about §5.1.2, or creating ambiguity around the term "document'), so let's not do that here.  Just referencing RFC3986 is sufficient.

PR #78, opened 2024-10-14
Align example section titles

### LINKED ISSUE #73: consistency on the naming of Overlay / Overlays
I think the official name is `Overlay` but there are quite a few references to `Overlays` (plural) in the specification and readme.md

If they are references to an Overlay description, I believe they should be lowercase or reworded.

https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/versions/1.0.0.md?plain=1#L137

https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/versions/1.0.0.md?plain=1#L165

https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/versions/1.0.0.md?plain=1#L188

https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/README.md?plain=1#L20

https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/README.md?plain=1#L22

ralfhandl:
Fixes #73

### DIFF
--- a/versions/1.0.0.md
+++ b/versions/1.0.0.md
@@ -134,7 +134,7 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 
 ### Examples
 
-#### Structured Overlays Example
+#### Structured Overlay Example
 
 When updating properties throughout the target document it may be more efficient to create a single `Action Object` that mirrors the structure of the target document. e.g.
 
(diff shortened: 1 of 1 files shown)


kevinswiber (changes requested):
The title in one of the examples, on line 172, still reads "Targeted Overlays". Can we update this to be the same as the Markdown title for the example? Otherwise, LGTM! Thank you!

mikekistler (approved):
Looks good. 👍

PR #87, opened 2024-10-21
1.x: the hopefully last two typos

ralfhandl:
Some minor copyedits that we spotted when reviewing around the time of the 1.0 release.

### DIFF
--- a/src/overlay.md
+++ b/src/overlay.md
@@ -197,10 +197,10 @@ actions:
   - target: $.paths.*.get.parameters[?@.name=='filter' && @.in=='query']
     update:
       schema:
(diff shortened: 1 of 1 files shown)


mikekistler (dismissed):
Looks good. 👍

lornajane:
Two things here:
 - This should be in a future patch release, so 1.0.1 or something like that, we won't update the 1.0.0 spec contents now.
 - This repo needs a branching strategy and documentation to go with it, and then we'll rebase this PR and use it as a target demo.

lornajane (approved):
First contributions to the 1.0.1 spec and they look good :)

PR #89, opened 2024-10-30
doc: Add apigee-go-gen CLI to the list tools that support Overlay 1.0

micovery:
The apigee-go-gen CLI (https://github.com/apigee/apigee-go-gen) now supports the Overlay 1.0 specification.

See docs at: https://apigee.github.io/apigee-go-gen/transform/commands/oas-overlay/


lornajane (approved):
Excellent! Happy to see more tools supporting the new standard, and thank you for adding it to the list :)

PR #94, opened 2024-11-05
feat: add schema tests and official JSON Schema

jeremyfiel:
includes pass and fail scenarios for Overlay schema tests. 

Also includes the full schema per #72


mikekistler (dismissed):
Looks good! 👍

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
+description: The description of OpenAPI Initiative Overlay v1.0.0 documents
+  without schema validation, as defined by
+  https://spec.openapis.org/overlay/v1.0.0
 type: object
+$comment: https://spec.openapis.org/overlay/v1.0.0#overlay-object
+properties:
+  overlay:
+    description: The version number of the Overlay Specification
+    type: string
+    pattern: ^1\.0\.\d+(-.+)?$
```
ralfhandl, line 12:
```suggestion
    pattern: ^1\.0\.\d+$
```

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
-$id: https://spec.openapis.org/overlay/1.0/schema/WORK-IN-PROGRESS
+$id: https://spec.openapis.org/overlay/1.0/schema/2024-10-17
 $schema: https://json-schema.org/draft/2020-12/schema
-description: The description of Overlay v1.0.x documents
-
+description: The description of OpenAPI Initiative Overlay v1.0.0 documents
+  without schema validation, as defined by
+  https://spec.openapis.org/overlay/v1.0.0
```
ralfhandl, line 5:
```suggestion
description: The description of Overlay v1.0.x documents
```
Following https://github.com/OAI/OpenAPI-Specification/blob/main/schemas/v3.0/schema.yaml

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
 type: object
+$comment: https://spec.openapis.org/overlay/v1.0.0#overlay-object
```
ralfhandl, line 7: Would we want to update all these `$comment`s when we publish v1.0.1?

I'd rather remove them all.

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
+properties:
+  overlay:
+    description: The version number of the Overlay Specification
```
ralfhandl, line 10: I'd rather remove all of these descriptions because they do not accurately reflect the specification text, which can/will cause confusion.

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
+required:
+  - overlay
+  - info
+  - actions
+$ref: "#/$defs/specification-extensions"
+unevaluatedProperties: false
+$defs:
+  info-object:
+    $comment: https://spec.openapis.org/overlay/v1.0.0#info-object
+    description: Provides metadata about the Overlay description
```
ralfhandl, line 33:
```suggestion
```

See above. Please apply to all other occurrences below.

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
+    unevaluatedProperties: false
+  action-object:
+    $comment: https://spec.openapis.org/overlay/v1.0.0#action-object
+    description: This object represents one or more changes to be applied to the
+      target document at the location defined by the target JSONPath expression.
+    properties:
+      target:
+        description: A JSONPath expression selecting nodes in the target document
+        type: string
+        pattern: ^\$[.\[]?.*$
```
ralfhandl, line 56:
```suggestion
```

Don't try to approximate the syntax defined in RFC9535 with a regular expression.

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,101 @@
...
+        if:
+          properties:
+            remove:
+              const: true
+          required:
+            - remove
+        then:
+          not:
+            required:
+              - update
```
ralfhandl, line 90: This is stricter than the specification, which allows `update` together with `remove: true`.

I'd rather remove this, and also remove the test case `fail/actions-no-update-with-remove.yaml`.

tests/v1.0/fail/actions-no-update-with-remove.yaml
ralfhandl, line 1: Remove this test case, see comment on `action-object`

ralfhandl (changes requested):

schemas/v1.0/schema.yaml
jeremyfiel: why diverge from OAS consistency?  This pattern is used in OAS and Arazzo.

schemas/v1.0/schema.yaml
jeremyfiel: just following OAS, I can remove them.

schemas/v1.0/schema.yaml
jeremyfiel: i asked for clarification on this in #71 

I think allowing this in the schema introduces an ambiguous implementation detail.. does the `remove` apply first, or the `update`?  Specifically because the spec indicates actions should be performed in order

schemas/v1.0/schema.yaml
ralfhandl: OAS needs this pattern because they published `3.0.0-rc0` etc. and `3.1.0-rc0` etc.

Arazzo doesn't need it because they only published `1.0.0` so far.

Consistently and blindly copying a too generous pattern is not a practice I embrace.

schemas/v1.0/schema.yaml
ralfhandl: Thanks for catching this, opened
* https://github.com/OAI/OpenAPI-Specification/issues/4181

schemas/v1.0/schema.yaml
ralfhandl: Let's wait for the outcome of #71 then with merging this PR.

ralfhandl: The schema validator we use thinks this `allOf` construct is invalid.

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,73 @@
 $id: https://spec.openapis.org/overlay/1.0/schema/WORK-IN-PROGRESS
 $schema: https://json-schema.org/draft/2020-12/schema
 description: The description of Overlay v1.0.x documents
-
-type: object
+properties:
```
ralfhandl, line 4:
```suggestion
type: object
properties:
```
Otherwise test case `not-an-object.yaml` will pass instead of fail

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,73 @@
...
+          - boolean
+          - object
+          - array
+          - number
+          - "null"
+      remove:
+        type: boolean
+        default: false
+    allOf:
+        if:
```
ralfhandl, line 57:
```suggestion
      - if:
```
`allOf` needs an array

ralfhandl, line 57:
```suggestion
    allOf:
      - if:
```
`allOf` is array-valued

schemas/v1.0/schema.yaml
```diff
@@ -1,5 +1,73 @@
...
+    required:
+      - title
+      - version
+    $ref: "#/$defs/specification-extensions"
+    unevaluatedProperties: false
+  action-object:
+    properties:
+      target:
+        type: string
+        pattern: ^\$[.\[]?.*$
```
ralfhandl, line 42: Given that `$` is a valid target, and `$foo` is not and matches the current pattern, we could shorten it to "start with `$`"
```suggestion
        pattern: ^\$
```

Or make it a bit more expressive as `^\$([.\[].*)?`

tests/v1.0/fail/extends-invalid-type.yaml
```diff
@@ -0,0 +1,6 @@
+overlay: 1.0.0
+info:
+  title: Invalid `extends` type
+  version: 1.0.0
+extends: {}
+actions: []
```
ralfhandl, line 6: Better add a valid action here to make sure that the `overlay` field causes the fail

tests/v1.0/fail/info-missing-title.yaml
```diff
@@ -0,0 +1,4 @@
+overlay: 1.0.0
+info:
+  version: 1.0.0
+actions: []
```
ralfhandl, line 4: Better add a valid action here to make sure that the `overlay` field causes the fail

schemas/v1.0/readme.md
```diff
@@ -0,0 +1,36 @@
+# OpenAPI Overlay 1.0.x JSON Schema
+
+Here you can find the JSON Schema for validating OpenAPI Overlay description of versions 1.0.x.
```
ralfhandl, line 3:
```suggestion
Here you can find the JSON Schema for validating Overlays of versions 1.0.x.
```
See [Style guide for Overlay Specification](https://github.com/OAI/Overlay-Specification/blob/main/CONTRIBUTING.md#style-guide-for-overlay-specification)

ralfhandl (commented):
Almost there, only minor nits.

Thank you!

ralfhandl:
@jeremyfiel Could you please add a `description` to one of the [Action Objects](https://spec.openapis.org/overlay/v1.0.0.html#action-object)?

That's the only part of the schema that is not yet tested.

jeremyfiel:
should be all covered now

ralfhandl:
> should be all covered now

Sorry, should have been more precise: please add to a "pass" test case, the "fail" ones don't count into the schema coverage.

jeremyfiel:
removed constraints for update/remove, updated tests

ralfhandl:
@jeremyfiel How can I view the diff between this state and the last state I reviewed?

jeremyfiel:
There is a `compare` hyperlink just above our comment box.  It's not available on mobile so I don't have it to share. It's available on desktop though

ralfhandl (approved):

mikekistler (approved):
Looks good! 👍

PR #95, opened 2024-11-08
schema test: improved error diagnostics

ralfhandl:
- error details if schema doesn't validate
- error details if "pass" test case fails


mikekistler (approved):
Looks good. 👍

PR #99, opened 2024-11-19
Add compliant example sets to show expected behaviour

lornajane:
These examples are taken from the test suite of https://github.com/lornajane/openapi-overlays-js (and with thanks to @MikeRalphson who helped with the tests originally!) to illustrate the expected outputs from applying particular Overlays to OpenAPI descriptions.

I think these examples could be improved/expanded, and I'm not sure if we should try to somehow standardise how whitespace/quotes work because they are most of the diff. Would it make sense to lint/format all the yaml? Advice welcome!


compliant-sets/README.md
```diff
@@ -0,0 +1,9 @@
+# OpenAPI Overlay Compliant Sets
```
ralfhandl, line 1:
```suggestion
# OpenAPI Overlay Compliance Tests
```
Or similar, and rename the top-level folder accordingly

ralfhandl:
> we should try to somehow standardise how whitespace/quotes work because they are most of the diff

I'd run Prettier on the YAML files to minimize these differences.

compliant-sets/README.md
lornajane: I think "Tests" rather overstates what these are: more like working examples.

compliant-sets/many-operations/output.yaml
```diff
@@ -0,0 +1,109 @@
...
+        default:
+          description: unexpected error
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/Error'
+      x-info:
+        x-overlay-applied: structured-overlay
+        description: This is an added description
+    post: &ref_0
```
ralfhandl, line 47:
```suggestion
    post:
```
I would not expect/require an Overlay processor to do this kind of YAML magic, I'd rather expect that it simply merges this everywhere, especially since the `post` operation may already exist and the Overlay would just add `foo` next to existing properties.

ralfhandl:
There should also be JSON inputs.

And some "mixed" inputs with OpenAPI JSON and Overlay YAML, and vice versa. Would the output format be determined by the format of the OpenAPI input? The spec doesn't tell.

compliant-sets/README.md
ralfhandl: I would expect that a compliant implementation produces equivalent output for the two inputs, and that implementors are well advised including these triplets in their end-to-end tests.

lornajane:
Fair enough. I don't think these are of a quality that we can publish for tests but from our conversation yesterday I thought we agreed that sharing whatever we had would be useful at this stage. Should I just close?

ralfhandl:
Don't close, I think these are a good starting point. Even if they aren't _complete_ yet, they are behavior that we expect from each implementation.

compliant-sets/many-operations/overlay.yaml
```diff
@@ -0,0 +1,23 @@
+overlay: 1.0.0
+info:
+  title: Structured Overlay
+  version: 1.0.0
+actions:
+- target: $.paths['/pets'].get.summary
+  remove: true
+- target: $.paths.*.get
```
ahx, line 8:
```suggestion
- $.paths.*.get.tags
```

Hey @lornajane, I have a question about this and the related output.yaml:

The spec says:

> If the target selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the target selects an array, the value of this field MUST be an entry to append to the array. 

The related part of the output expects "wildcard-done" to be added to the list of tags. But the target ($.paths.*.get) does not select an array. So should this be `$.paths.*.get.tags` instead?

ahx:
This change is super useful, even in it's current state. I am submoduling this right now and building a ruby library against this ♥️

compliant-sets/README.md
lornajane: Tests would be overstating it, I think we need to stay with vague language here.

compliant-sets/many-operations/output.yaml
lornajane: I've removed this example from the pull request for now.

lornajane:
@ahx thanks for the encouraging words! I've updated the PR and re-requested reviews for it.

mikeschinkel:
This looks really great and would really like to see it approved.

compliant-sets/many-operations/overlay.yaml
ahx: Sorry, I should probably just ask somewhere else about how to read the spec. The spec says "merge", which I understand like `Object.assign(a, b) `  in JavaScript, where if properties defined in a will be overwritten by the same properties defined b. Array values would just be overwritten and not concatenanted.

Is my understanding wrong?

ralfhandl (changes requested):
Can we normalize all YAML files so that there are no formatting variations (single/double/no quotes, empty lines, ...) and the diff between `openapi.yaml` and `output.yaml` is just the changes made by `overlay.yaml`?

PR #103, opened 2024-11-26
Generate minor-version symlinks

ralfhandl:
Merge recent changes from OpenAPI and Arazzo publishing workflows:
* `latest.html` is now a symlink instead of a copy
* `v{MAJOR}.{MINOR}.html` symlink that could be referenced in schema YAML file(s)


lornajane (approved):
Did what I expected to when tested locally, thank you for adding these!

PR #105, opened 2024-11-26
feat(linting): add markdownlint to all readme files

### LINKED ISSUE #104: Consider markdown lint on all .md files (spec, schema readme)
clean up markdown formatting on all *.md files, not only the spec.

jeremyfiel:
fixes #104


lornajane:
Let's add separate linting standards for internal README files and the official specification versions. If you want to add some linting on README files that's fine, but the requirements for these two content types aren't common enough and I expect us to make more and stricter rules for the specifications at some point.

PR #109, opened 2024-12-04
schema test coverage: return code 1 if coverage is below 100%

### LINKED ISSUE #107: schema-tests workflow: fail if coverage drops below 100 percent
- [ ] extend [schema-test-coverage.mjs](https://github.com/OAI/Overlay-Specification/blob/main/scripts/schema-test-coverage.mjs) with return code 1 if schema parts are uncovered 
- [ ] extend [schema-test-coverage.sh](https://github.com/OAI/Overlay-Specification/blob/main/scripts/schema-test-coverage.sh) to propagate return code

ralfhandl:
This will cause the schema-test workflow to fail if the schema test coverage drops.

Fixes #107

### DIFF
--- a/scripts/schema-test-coverage.mjs
+++ b/scripts/schema-test-coverage.mjs
@@ -121,6 +121,7 @@ if (notCovered.length > 0) {
   const firstNotCovered = notCovered.slice(0, maxNotCovered);
   if (notCovered.length > maxNotCovered) firstNotCovered.push("...");
   console.log(firstNotCovered);
+  process.exitCode = 1;
 }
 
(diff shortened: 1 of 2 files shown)


lornajane:
Adding some notes as I go along:
 - the action that runs this is `schema-tests`
 - that in turn runs `npm test`
 - tested it by removing enough tests to reduce the coverage number and confirming that `npm test` on the main branch returns exit code 0 but on this branch I get exit code 1 as described.

lornajane (approved):
Thank you, a small but mighty change!

PR #113, opened 2024-12-22
Workflows with app-generated access token

ralfhandl:
Instead of using personal access tokens the two workflows for publishing HTML specs and JSON Schemas now use an app-generated temporary access token.

Same changes as in
* https://github.com/OAI/Arazzo-Specification/pull/302

Example result 
* https://github.com/OAI/OpenAPI-Specification/pull/4277


lornajane (approved):
Thanks for getting this sorted out!

PR #115, opened 2025-01-01
Use multiple markdownlint.yaml /versions and /schemas

Bellangelo:
Closes: https://github.com/OAI/Overlay-Specification/issues/104

Here it was decided to use different markdownlint.yaml files for `/schemas` and `/versions` since they have different purposes and as such, different standards.

What this PR introduces:
1. A `markdownlint.yaml` exist in each folder, `/versions` and `schemas`. Sub-folder linting seems to be supported by most plugins. See https://github.com/DavidAnson/vscode-markdownlint/issues/280
2. `format-markdown.sh` requires now a `--config` parameter since now the config is not in a single place.
3. Renamed and group our scripts in the `package.json` to use the same namespaces.


schemas/.markdownlint.yaml
```diff
@@ -0,0 +1,14 @@
...
+MD012: false # allow blank lines
+
+MD013:
+    line_length: 800
+    tables: false
+
+MD024: false # duplicate headings
+MD032: false # blanks-around-lists Lists should be surrounded by blank lines
+MD033: false # inline HTML
+MD047: false # single-trailing-newline Files should end with a single newline character
```
Bellangelo, line 14: Rules 32 and 47 are disabled because they didn't pass. We can enable them again and format the file if that is what we desire.

ralfhandl:
> it was decided to use different markdownlint.yaml files for `/schemas` and `/versions`

Actually we'd like to apply the same rules to all Markdown files in this repository, including the ones in the repository root.

schemas/.markdownlint.yaml
ralfhandl: Yes, please use the same rules for all `*.md` files and change the files that don't follow the rules yet.

scripts/format-markdown.sh
```diff
@@ -2,7 +2,28 @@
...
+      ;;
+  esac
+done
+
+if [ -z "$CONFIG" ]; then
+  echo "Error: --config parameter is required."
+  exit 1
+fi
+
+for filename in $FILES; do
```
ralfhandl, line 26: Not necessary, same rules for all files

versions/.markdownlint.yaml
ralfhandl, line 1: Please move back to root folder

ralfhandl (changes requested):

schemas/.markdownlint.yaml
lornajane: I'm not sure I agree that the same rules apply for specification documents and our internal project documentation in markdown files. I think we already rejected a pull request that applied the same standards to both - We need to be strict and careful with the documents that are published, I'm not sure the same applies if people are adding different content types.

Bellangelo: HI @ralfhandl , I am not against in what you are proposing but @lornajane is right about the rejected PR. Please check https://github.com/OAI/Overlay-Specification/pull/105

ralfhandl: Given that the commented-out rules are trivial to observe, I'd use the same rules. 

The more interesting case will be when we make the rules for new spec documents stricter and will need different rule sets for files in the `versions` folder, as we already do for OAS.

Bellangelo: @ralfhandl Ok. But what this means for the other comments that you made? Such as https://github.com/OAI/Overlay-Specification/pull/115#discussion_r1900669324

To be more precise, are we keeping the multiple lint configs or are we using a single one?

ralfhandl: I prefer keeping a single rule file and adjusting all Markdown files across all folders accordingly.

lornajane (commented):
Apparently I didn't finish my review with comment!

versions/.markdownlint.yaml
lornajane: I'd advocate for locating a config file with a stricter set of rules here, with the all-files rules at the root of the project. We need to use the strict rules for `src/overlay.md` file on the development branches as well. However I don't see any markdown files in the `schemas/` directory so I'm not sure why we have put configuration there.

lornajane:
It took me a while to come back to look at this, sorry.

My suggestion:
 - a basic set of markdownlint configuration in the root folder that's used for everything except the specification itself
 - a second markdownlint configuration used for `versions/` and `src/` files so that we can be much more precise about having good standards there, which we don't need for our README files.

scripts/format-markdown.sh
lornajane: I'm not sure what this is used for but I definitely don't see that we need this script, can someone explain what the problem was that this was solving? Let's try again!

lornajane:
Here's the related change on the main OpenAPI repo https://github.com/OAI/OpenAPI-Specification/pull/4515. I'm going to close this pull request for now (sorry @Bellangelo, I really appreciate the discussion and improvements that your work provoked on this!)

PR #118, opened 2025-01-29
Fixes examples for compliance with RFC-9535

mikekistler:
This PR makes a few minor fixes to the overlay examples to ensure they are compliant with RFC-9535.

I tested all the examples in the spec with Speakeasy's overlay playground and only these two had problems.

Note that any examples that use filters need
```yaml
x-speakeasy-jsonpath: rfc9535
```

but I did not include that in the examples.


versions/1.0.0.md
```diff
@@ -252,15 +252,16 @@ info:
   title: Apply Traits
   version: 1.0.0
 actions:
-  - target: $.paths.*.get[?@.x-oai-traits.paged]
+  - target: $.paths[?@.get && @.get['x-oai-paged']]
```
gregsdennis, line 255: The `@.get` on its own isn't necessary.
```suggestion
  - target: $.paths[?@.get['x-oai-paged']]
```

lornajane (approved):
Thanks!

kevinswiber:
@mikekistler @gregsdennis Will this query work if we keep the original `x-oai-traits` example?

```
$.paths[?@.get['x-oai-traits'][?@ == 'paged']]
```

```yaml
openapi: 3.1.0
info:
  title: API with a paged collection
  version: 1.0.0
paths:
  /items:
    get:
      x-oai-traits: ['paged']
      responses:
        200:
          description: OK
```

mikekistler:
Thanks @kevinswiber and @gregsdennis for puzzling out how to use the original x-oai-traits with array value. I've updated the PR to incorporate this.

src/overlay.md
```diff
@@ -212,8 +212,8 @@ info:
 actions:
   - target: $.paths.*.get.parameters
     update:
-      name: newParam
-      in: query
+      - name: newParam
+        in: query
```
ralfhandl, line 216: The spec says
> If the target selects an array, the value of this field MUST be an entry to append to the array.

The original example appended a Parameter Object to an array of Parameter Objects, which feels correct.

The modified example appends a one-element array to an array, so I would expect the result be an array whose last element is the one-element array.

If we want "append to an array" to mean "concatenate two arrays", we should make that really clear in the text.

ralfhandl (changes requested):

src/overlay.md
kevinswiber: @ralfhandl I don't think we do want that. I think the behavior you described is what I would expect. We had a whole issue and PR discussing arrays. I think it's #30?

lornajane: @ralfhandl can you take another look at this? Do we need to make the changes you requested?

lornajane (approved):
I'm happy to approve this but we still have some requested changes to look at.

src/overlay.md
ralfhandl: Well, that depends 😎 

The Overlay specification says for [action field `update`](https://spec.openapis.org/overlay/v1.0.0.html#fixed-fields-1)
> If the target selects an array, the value of this field MUST be an entry to append to the array.

I would interpret this as
* If I want to add a Tag Object to the `tags` array, the value of `update` must be single the Tag Object I want to add.
* If I want to add two entries, I need two actions, each of which appends one entry to the array.

This interpretation is supported by the first [Array Modification Example](https://spec.openapis.org/overlay/v1.0.0.html#array-modification-example) in the 1.0.0 spec.

However, both [Bump.sh](https://github.com/bump-sh/cli) and [Speakeasy](https://www.speakeasy.com/openapi/overlays#adding-tags-and-updating-drink-responses) require `update` to be an _array_ with one or more entries, and this array is _concatenated_ with the target.

If we want to keep the spec as it is now, the array modification example must not be changed, and we live with the fact that the two first implementations of the spec got array modification wrong.

If we want to make the example compatible with the two first implementations, we also must change the specification and say something like

> If the target selects an array, the value of this field MUST be an <ins>array of entries</ins><del>entry</del> to append to the array.

or

> If the target selects an array, the value of this field MUST be an <ins>array</ins><del>entry</del> to <ins>concatenate with</ins><del>append to</del> the <ins>target</ins> array.

This however would be a breaking change and should be done in a new major version of the spec.

versions/1.1.0-dev.md
```diff
@@ -212,8 +212,8 @@ info:
 actions:
   - target: $.paths.*.get.parameters
     update:
-      name: newParam
-      in: query
+      - name: newParam
+        in: query
```
ralfhandl: This should not be a one-element array. [According to the spec](https://spec.openapis.org/overlay/v1.0.0.html#action-update) it should be an object

> If the target selects an array, the value of this field MUST be an entry to _append_ to the array.

Appending a one-element array to an array results in a nested array.

Both Overlay implementations got that wrong and used `concat()` instead of `push()`.

We cannot change the spec in a minor version to suit the implementations.

The implementations probably have similar concerns.
```suggestion
      name: newParam
      in: query
```

ralfhandl: See outdated conversation on `src/overlay.md` above.

mikekistler: Okay ... I've backed out this change.

If there are tools that aren't compliant with the spec, we should:
- open an issue in their repo
- remove them from the list of compliant implementations in the README.

But we can do that in another PR.

versions/1.1.0-dev.md
```diff
@@ -252,7 +252,7 @@ info:
   title: Apply Traits
   version: 1.0.0
 actions:
-  - target: $.paths.*.get[?@.x-oai-traits.paged]
+  - target: $.paths[?@.get['x-oai-traits'][?@ == 'paged']].get
```
ralfhandl: [Playing with bump-cli and speakeasy](https://github.com/ralfhandl/overlay-experiments) I found that this works:

```suggestion
  - target: $.paths.*.get[?(@.x-oai-traits[?(@ == 'paged')])]
```

Both tools require the parens after the `?` although [RFC9539](https://www.rfc-editor.org/rfc/rfc9535#tbl-filter) has 25 examples without parens and just one with parens to tell that expressions with parens are equivalent to ones without 😁

gregsdennis: Tools probably aren't up to spec yet, especially with the prevalence of libs like jsonpathplus, which extend classic JSON Path.

ralfhandl (approved):

versions/1.1.0-dev.md
```diff
@@ -252,7 +280,7 @@ info:
   title: Apply Traits
   version: 1.0.0
 actions:
-  - target: $.paths[?@.get['x-oai-traits'][?@ == 'paged']].get
+  - target: $.paths.*.get[?(@['x-oai-traits'][?(@ == 'paged')])]
```
gregsdennis, line 283: This change isn't necessary for RFC-9535, but it would increase the compatibility with legacy libraries.

Issue #33, opened 2017-08-31
Environment Variables in definition


Download:
Sorry if this came up before, I searched but could not find anything.

Environment variables have become something of the de facto mechanism to set server-specific values, then access them from software. The hosting providers I have worked with (AWS, Heroku, OpenShift) all offer features to easily set/import/export them and most software I've seen meant to run on these systems uses it for things like environment specific URLs, secrets etc.

But afaik there is no way to reference environment variables in an OpenAPI definition. 

Would this be a feature to support? I imagine the definition having a section (probably at the top level) that specified a set of environment variables expected to be set, and possibly the default values to use when they are not set. Those variables could then be used in other parts of the definition.

My practical use cases comes from needing to set different authorization urls for different environments. Currently I am rendering swagger.json from JSP to implement this myself, but it seems to me this is a very generic use case that many people will face.

See also e.g.
https://github.com/OAI/OpenAPI-Specification/issues/169
https://github.com/OAI/OpenAPI-Specification/issues/779
https://github.com/swagger-api/swagger-ui/pull/3410

lornajane:
Being able to use an environment variable as a value to use in an update would cover quite a few of these use cases. How about using a `$something` [syntax like Arazzo](https://spec.openapis.org/arazzo/v1.0.1.html#examples) does for `$inputs` and `$response` and things? So we would end up with something like:

```
overlay: 2.0.0
info:
  title: Use Env Vars to add some extra data
  version: 1.0.0
actions:
  - target: $.paths.*.get
    update:
      x-extra: $env.EXTRA_FIELD_VALUE
```

There might be other sources of replacement values so using `$env` makes it clear where this value comes from, and leaves room for more options (like `$target` or `$doc` or .... some other things I haven't thought of it yet).

ralfhandl:
I like the idea, and I am a bit surprised by the matrix feature:
- Which use cases do you have for it? In the example the last update wins and the matrix has no effect. 
- Will it also work with environment variables if I `export foo=("one" "two")` before applying the overlay?

baywet:
I think it could also work with environment variables if we defined a syntax for "arrays" essentially.

The main use case we have for it is that we have a bunch of component schemas we want side cars for.
Right now, our overlay that I can't share publicly, has 40+ actions just for those side car definitions. With a clever JSONPath definition and this matrix ability, we could get it down to 1, making the maintenance significantly easier.

baywet:
[...]

PR #140, opened 2025-07-10
feat: add capability to use existing content of OpenAPI as input to allow for renaming/editing scenarios

### LINKED ISSUE #121: Capability to copy (deep) nodes from one target to another target
The [Actions Object](https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md#action-object) has the ability to enable users to add/mutate/remove attributes on OpenAPI documents, but cannot handle the following scenarios:

1. Renaming `paths` items

I.e. given a path on `$.paths["/item"]`, move it to `$.paths["/newitem"]:

From

```
paths:
  /item:
    summary: 'The root resource'
    get:
      summary: 'Retrieve the root resource'
      ...
```

To:

```
paths:
  /newitem:
    summary: 'The root resource'
    get:
      summary: 'Retrieve the root resource'
      ...
```

Doing this via a full add/remove of the `/item` requires inlining the schema into the overlay, and thus makes it brittle: any overlay which overrides a path item in this way effectively nukes the ability for a source document to change any of the information in "/item".

2. Extracting inline schemas into component schemas

Many OpenAPI frameworks produce inline json schemas for operations:

```
paths:
  /item:
    post:
      requestBody:
        required: true
        content:
          application/json:
            schema:
              type: object
              required:
                - name
                - photoUrls
              properties:
                id:
                  type: integer
                  format: int64
                name:
                  type: string
```

Enabling an overlay to extract that inline schema and replace it with `$ref: #/components/schemas/Item` would enable a category of simplifications for downstream tooling.

# Proposed change to the Actions Object

##### Fixed Fields

| ---- | :----: | ---- |
| <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. |
| <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object or array MUST be removed from the the map or array it is contained in. The default value is `false`. |
| <a name="action-destination"></a>destination | `boolean` | A JSONPath expression selecting destination nodes in the target document. If set, this indicates that each node in the target expression (after applying the update) should be appended as a child to the destination node. If not set, the target is mutated in-place. |

ThomasRooney:
Closes #121


mikeschinkel:
Today @lornajane and I discussed this on call with both of us finding the exact syntax proposed not idea. 

@lornajane took the position that `update` should be enhanced to support this used case, but that would necessitate a `2.0.0` release. She wanted to enhance `update` to make it more powerful so that it could address this use-case as well as others _(@lornajane please correct my characterization if I got it wrong.)_

My perspective is that we could instead offer a `rename` property in addition to `update` that could allow for keeping same major version — i.e. no backward compatible breakage — and having a `rename` would make it more obvious to those looking for such a feature.  It would also decouple the use-cases so that `update` would not have to be compatible with a rename use-case, existing uses-cases, and future use-cases; `rename` could instead be bespoke for its needs and only its needs.

A quick strawman example might look like this _(and since I am doing this in a rush before the next meeting I will most likely not get this right, but hopefully I will.):_

```
overlay: 1.0.0
info:
  title: Rename path items
  version: 1.0.0
actions:
  - target: $.paths
    update: 
      "/newitem": {}
  - target: $.paths
    rename:
      existing: /items
      new: /newitems
    remove: true
```    

Please comment and bikeshed away.

lornajane:
Perhaps I did not explain myself well although I did try in the discussion. I suggested that we could support the use case that this pull request addresses by adding [support for input sources](https://github.com/OAI/Overlay-Specification/issues/33#issuecomment-2936023796). This does not require a major version bump, unless we need to add extra options for handling content such as `$` signs within OpenAPI project (since the proposed new syntax would support `$target` and `$doc`) - the additional options probably want to go inside `update` and that change would cause a major version bump. Mike suggested adding additional operations with their own data types for each supported use case, so I did not proceed with the work to propose the additional input sources.

However we have both thoroughly invaded @ThomasRooney 's original pull request - I'll add some comments in line with the changes.

src/overlay.md
```diff
@@ -115,8 +115,9 @@ This object represents one or more changes to be applied to the target document
 | ---- | :----: | ---- |
 | <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
 | <a name="action-description"></a>description | `string` | A description of the action. [[CommonMark]] syntax MAY be used for rich text representation. |
-| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true`. |
```
lornajane, line 118: Why change this behaviour? Remove will still cause the other fields to have no impact?

src/overlay.md
```diff
@@ -115,8 +115,9 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object or array MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-destination"></a>destination | `string` | A JSONPath expression selecting destination nodes in the target document. If set, this indicates that each node in the target expression (after applying the update) should be appended as a child to the destination node. If not set, the target is mutated in-place. |
```
lornajane, line 120: I support solving the use case, but I am not in favour of this syntax. In 1.0, the changes are applied to the nodes that match the `target` expression. In this instance, the target isn't the target, but is more like the source which has potential to be confusing.

Is there another way we could indicate which field to update, and where to get the data from to put there?

src/overlay.md
```diff
@@ -226,6 +227,51 @@ actions:
...
+info:
+  title: Rename path items
+  version: 1.0.0
+actions:
+  - target: $.paths
+    update: 
+      "/newitem": {}
+  - target: $.paths["/item"]
+    destination: $.paths["/newitem"]
+    remove: true
```
lornajane, line 249: I'd prefer to see copy and remove as two different steps - it's likely that we'll do `action` as `update` or `remove` (or whatever else) in a future version and I think it's clearer.

PR #148, opened 2025-08-21
adds a schedule for the bi-weekly agenda discussion creation

baywet:
the workflow will now trigger every other week (when we don't meet) and create the meeting for the next week.
Since the next agenda has already been created manually, I suggest that after merge we DISABLE the workflow, and only re-enable it on 2025-09-01 so it's in sync with the schedule we have.


lornajane (approved):
Thank you!

PR #150, opened 2025-10-14
Adds a copy action

### LINKED ISSUE #121: Capability to copy (deep) nodes from one target to another target

### LINKED ISSUE #146: Proposal - add support for sourcing an update value from another node of the document
Hey everyone,
Thank you for putting this together, overlays are extremely helpful for cleaning up OAI descriptions owned by third parties.

One scenario we can across multiple times is the use of `oneOf` to specify nullability with something like this:

```YAML
components:
  schema:
    Foo:
      oneOf:
        - type: string #using string here for simplicity, but you could imagine a complex object type being described instead
        - nullable: true
```

In our scenario, we're importing OAI descriptions to mechanically process them down the road (doc generation, code generation, etc...). When the tool downstream sees that pattern it "jumps to conclusions" and emits a union type. Although that's technically correct, it's incorrect in principle. We could of course change that tool and add a bunch of specialization conditions to say "sure, but if you see that pattern, it's not really a union type".

To clean it up before processing, we end up with an overlay that looks something like this

```json
{
  "actions": [
    {
      "update": {"type": ["string", "null"]},
      "target": "$.components.schemas['Foo']"
    },
    {
     "target": "$.components.schemas['Foo'].oneOf",
     "remove": true
    }
```

Although that *works*, it leads to a lot of duplication we have to maintain (again, imagine that instead of type string, you have a complex object definition with a bunch of properties, etc...)

We'd much rather have the following overlay instead.

```json
{
  "actions": [
    {
      "update-from-source": "$.components.schemas['Foo'].oneOf.[? (@.type == 'string')]",
      "target": "$.components.schemas['Foo']"
    },
    {
     "target": "$.components.schemas['Foo'].oneOf",
     "remove": true
    },
    {
     "target": "$.components.schemas['Foo']",
     "update": {"type": ["string", "null"]}
    }
```

A couple of notes on that example:
- yes, it's technically more actions, but now the **only** thing I have in the update, is the type field, no duplication for the rest of the schema.
- I gave an ugly name to that new "update-from-source" new action kind so we can give it a better name together.
- This new action would effectively behave like update behaves, but source the update value to merge into the target using a JSON Path instead of a "hardcoded" value.
- I think we'll need to define the behaviour as "source the update value first, and entirely, then apply the update" to prevent infinite recursion kind of problems.

If this new thing is desirable, I'm happy to draft something up, and potentially have it released as part of a 1.1 version of the spec.

baywet:
fixes #146 
fixes #121
closes #140

Hi everyone,
I've tried to get to the simplest common denominator, and incorporate some of the feedback that was already provided:
- We want the ability to use what's already in the document to update the document.
- We want single purpose actions, that are mutually exclusive, atomic, and can be combined like lego bricks.
- We want our "action keywords" to be verbs.

I'm proposing to add a **copy** field to the action, that's mutually exclusive with remove or update. It's functionally equivalent to the update field except that it uses a JSON path to source the JSON Node instead of defining it inline.

I know there was some additional thinking about templating expressions, sourcing from env variables etc... But I think those can be considered orthogonal and be added in a separate PR by their champions. I'd like to get the smallest possible unit that brings a meaningful contribution through.

As an image is better than a thousand words, here are a couple of examples.

## Copying a node to a node that already exists

```yaml
overlay: 1.1.0
info:
  title: Merge an existing path item with another one
  version: 1.0.0
actions:
  - target: '$.paths["/bar"]'
    copy:  '$.paths["/foo"]'
```

Here the existing foo path item properties are being copied to the bar path item.

## Copying a node to a node that DOES NOT already exists

```yaml
overlay: 1.1.0
info:
  title: Merge an existing path item with another one, ensuring the target exists
  version: 1.0.0
actions:
  - target: '$.paths'
    update: { "bar": {} }
  - target: '$.paths["/bar"]'
    copy:  '$.paths["/foo"]'
```

## Moving an item

```yaml
overlay: 1.1.0
info:
  title: Move an existing path item to another one
  version: 1.0.0
actions:
   # optionally add a remove action to ensure "a clean slate" for the bar path item
  - target: '$.paths'
    update: { "bar": {} }
  - target: '$.paths["/bar"]'
    copy:  '$.paths["/foo"]'
  - target: '$.paths["/foo"]'
    remove: true
```

A couple of todos if that gets consensus:
- [x] We need to stand up a v1.1-dev branch, and this needs to be rebased
- [x] We need to add an example in the spec
- [x] We need to update the schema
- [x] We need to add some tests
- [ ] Editorial information (version, table, etc...) need to be applied

### DIFF
--- a/versions/1.1.0-dev.md
+++ b/versions/1.1.0-dev.md
@@ -272,7 +272,7 @@ This approach allows inversion of control as to where the Overlay updates apply 
 
 #### Copy example
 
-Copy actions behave similarly to `update` actions but source the new values to merge with the target from the document being transformed instead of providing them in the Overlay document. Copy `actions` MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
+Copy actions behave similarly to `update` actions but source the new values to merge with the target from the document being transformed instead of providing them in the Overlay document. Copy actions MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
 
 ##### Simple copy
 


ThomasRooney (approved):
I personally really like this. It's simple nomenclature and implementation.

Annecdotally to me the alternative approach of using runtime expressions is more powerful in some ways, but also limiting in others. We've been using them at Speakeasy to represent complex concepts in Arazzo but unfortunately we've had to build escape hatches for places where they don't work (specifically the lack of conditionals). I'd propose we accept this as a minor version bump, and do that as an independent addition to this change as part of future work given a set of differing use-cases.

src/overlay.md
```diff
@@ -114,8 +114,9 @@ This object represents one or more changes to be applied to the target document
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
+| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting locations to copy target nodes to in the target document. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. If the `target` selects multiple nodes, the values will be copied sequentially with each node selected by the `copy` expression. |
```
mikekistler: I'm not clear on what this means:
> If the `target` selects multiple nodes, the values will be copied sequentially with each node selected by the `copy` expression.

It would be helpful to have an example of this. And I think this and all of the examples currently in the PR description should be included in the spec.

baywet: what I was trying to express here is that:
1. the number of nodes the target and the copy expressions match must be identical
2. should their be more than one match, the copy match 1 will apply to target match 1, then copy match 2 will apply to target match 2, etc...

But I couldn't come up with a succinct way of saying this

baywet: I've tried to feed as much context to copilot as possible and here is what it came back with as a replacement.

> When both the target and copy expressions match multiple nodes, the cardinality of the matches MUST  be equal. Each copy node is applied to the corresponding target node in positional order: the first copy node to the first target node, the second to the second, and so forth.

What do you think?

baywet: (updated, based on the thumbs up)

tests/schema.test.mjs
```diff
@@ -30,38 +30,35 @@ const parseYamlFromFile = (filePath) => {
...
 };
 
+const runTestSuite = (version, validateOverlay, suite = "pass") => {
+  readdirSync(`./tests/v${version}/${suite}`, { withFileTypes: true })
+    .filter((entry) => entry.isFile() && /\.yaml$/.test(entry.name))
+    .forEach((entry) => {
+      test(entry.name, () => {
+        const instance = parseYamlFromFile(`./tests/v${version}/${suite}/${entry.name}`);
+        const output = validateOverlay(instance, BASIC);
+        expect(output.valid).to.equal(suite === "pass");
```
ralfhandl: If a "pass" test case fails, this test code will not provide any info on why it failed. That was the reason for 

~~~js
expect(output).to.deep.equal({ valid: true });
~~~

The validation output contains details on what failed validation next to the `valid: false` property which shows up in the test log.

Better use something like
```suggestion
        if (suite === "pass")
          expect(output).to.deep.equal({ valid: true });
        else
          expect(output.valid).to.equal(false);
```

tests/v1.0/pass/actions-description.yaml
```diff
@@ -5,3 +5,4 @@ info:
 actions:
   - target: '$' # Root of document
     description: this is an action description
+    update: {}
```
ralfhandl: Why change v1.0 test cases in a v1.1 branch? These should have passed the v1.0 schema.

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,53 @@ actions:
...
+
+```yaml
+overlay: 1.1.0
+info:
+  title: Demonstrates variations of "copy" uses
+  version: 1.0.0
+actions:
+  - target: '$.paths["/some-items"]'
+    copy: '$.paths["/items"]'
+    description: 'copies recursively all elements from the "items" path item to the new "some-items" path item without ensuring the node exists before the copy'
```
ralfhandl, line 311: What happens if the target does not exist?

src/overlay.md
```diff
@@ -114,8 +114,9 @@ This object represents one or more changes to be applied to the target document
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="action-target"></a>target | `string` | **REQUIRED** A JSONPath expression selecting nodes in the target document. |
+| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting locations to copy target nodes to in the target document. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. When both the target and copy expressions match multiple nodes, the cardinality of the matches MUST be equal. Each copy node is applied to the corresponding target node in positional order: the first copy node to the first target node, the second to the second, and so forth. |
```
ralfhandl: What if the target selects nothing?

tests/v1.0/pass/actions-description.yaml
baywet: At some point I had changed the test definitions to also run the v1.0 files for v1.1. I can revert those if we think that's an issue.

versions/1.1.0-dev.md
baywet: In case the target does not exist, same as remove/update, NO-OP
In case the copy does not exist, we could define this as NO-OP, or error? I'm happy either way.

src/overlay.md
baywet: (let's keep the conversation in the other thread until we come to a decision, and I'll revert back here to update the text once we have)

versions/1.1.0-dev.md
```diff
@@ -126,6 +127,8 @@ Primitive-valued items of an array cannot be replaced or removed individually, o
 
 The properties of the `update` object MUST be compatible with the target object referenced by the JSONPath key. When the Overlay document is applied, the properties in the `update` object are recursively merged with the properties in the target object with the same names; new properties are added to the target object.
 
+The properties of the resolved `copy` object MUST be compatible with the target object referenced by the JSONPath key. When the Overlay document is applied, the properties in the resolved `copy` object are recursively merged with the properties in the target object with the same names; new properties are added to the target object.
```
RobertCraigie, line 130: question: what does "compatible" mean here? I don't see it defined more concretely elsewhere in the spec?

src/overlay.md
lornajane: I'd put this field in between `update` and `remove` in the table. All actions need `target` and `description`, so it's a bit weird to split them.

src/overlay.md
```diff
@@ -265,6 +268,53 @@ actions:
 
 This approach allows inversion of control as to where the Overlay updates apply to the target document itself.
 
+#### Copy example
+
+Copy actions behave similarly to update actions but source the node to from the document being transformed. Copy actions MAY be combined with update or remove actions to perform more advanced transformations like moving or renaming nodes.
```
lornajane: Perhaps "used in sequence with" rather than "combined with" since the fields are mutually exclusive and can't be combined - so this could be confusing.

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,53 @@ actions:
 
 This approach allows inversion of control as to where the Overlay updates apply to the target document itself.
 
+#### Copy example
```
lornajane, line 273: I would like to see a sentence before and after each code snippet explaining what is being illustrated in the snippet, and the outcome of it.

baywet:
> Really I would like the "make a 1.1-dev branch" changes in a different pull request from the "add a copy action" changes. They are different things.

@lornajane Are you referring to the unit tests infrastructure changes? Or would you like me to target a different branch all together?

versions/1.1.0-dev.md
RobertCraigie: What's the reasoning for making it a no-op/error instead of adding the node for you? Even if that just works for the direct target, and doesn't create parent nodes. (sorry if this is in the spec, I couldn't find it)

I'm asking as for our versions of "copy" transformations, we create the target node for you. Although admittedly our version is special cased for copying paths and schema objects, so creating the target node is very simple as we don't have to even define what the behaviour would be for recursively creating parent nodes.

I do think that creating the node if it didn't already exists would be pretty helpful as currently you would need to define two different actions with a _different_ path syntax, instead of just one action, or even two actions with the same path.

Although this seems like a fundamental aspect of the Overlays design, so it probably does not make sense to try and figure this out just for `copy`.

versions/1.1.0-dev.md
RobertCraigie, line 130: > are recursively merged with the properties in the target object

fwiw the deep merge behaviour here is surprising to me, generally I'd expect a "copy" operation to behave similarly to a spread in typescript with `{ ...original, ...newValues }` and I'd expect the deep merge behaviour for an operation named `merge`, `deep-merge`, or `deep-copy`.

I'm curious what the motivating use cases are for deep merging?

RobertCraigie (commented):
Thanks for the ping @mikekistler! Super excited to see this is being worked on :)

I'll note that as someone that is not deeply familiar with Overlays yet, on first glance it was not immediately clear to me which json path was the one being updated, but it did click very quickly after I actually thought about it.

It feels silly to write that out as it is pretty clear now but that was my initial reaction, so I'm curious if you considered another name for the operation to be super explicit? e.g. `copy-from`?

tests/v1.0/pass/actions-description.yaml
baywet: (resolving this thread, we can address any further issues in #180 and I'll rebase here)

src/overlay.md
baywet: updated.

versions/1.1.0-dev.md
baywet: added sections, and resulting descriptions to all examples.

src/overlay.md
baywet: updated

versions/1.1.0-dev.md
baywet: For the target: I'm happy either way, I just want the behavior to be consistent with other kinds of actions. As far as I understand the current spec (v1.0), if the target yields 0 matches, the behavior is currently implementation specific (could be a NO-OP, could be a warning, could be an error). And changing that would probably represent a breaking change from a behavior standpoint. Please correct me if I missed anything.

As for the copy, as written right now, if the cardinality does not match what target found, it's an error. I think that's probably the safest behavior for implementers. There's no point erroring out when copy finds nothing, if at the same time the target found nothing. We might want to rephrase things to we don't need to error when copy finds more items than target.

versions/1.1.0-dev.md
baywet: I used the same linguo as for update for consistency. As far as I understand it means same object type/shape (by that I mean OpenAPI object)

baywet: Again, this is written in consistency with the update keyword. If we want to change that, it should be changed for both for consistency. And changing it for update most likely represents a behavior breaking change at this stage.
If we want additional keywords that impact the copy/update merge behavior, I'm not against it, but this should probably be considered outside of this scope to avoid scope creep and submitted as a separate pull request.

src/overlay.md
```diff
@@ -115,7 +115,8 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
+| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting locations to copy target nodes to in the target document. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. When both the target and copy expressions match multiple nodes, the cardinality of the matches MUST be equal. Each copy node is applied to the corresponding target node in positional order: the first copy node to the first target node, the second to the second, and so forth. |
```
baywet: from the meeting: restrict copy to have a single source so it's better aligned with update and potentially less brittle. Iterate based on feedback after new version release.

versions/1.1.0-dev.md
baywet: from the meeting, compatible is not explicitly defined, but this should be done as a separate PR, let's create a separate issue for that.

schemas/v1.1/schema.yaml
```diff
@@ -0,0 +1,75 @@
...
+    unevaluatedProperties: false
+  action-object:
+    properties:
+      target:
+        type: string
+        pattern: ^\$
+      description:
+        type: string
+    oneOf:
+      - properties:
```
baywet: revert from oneOf to a set of properties, because it's explicitly allowed in the spec even though it's semantically meaningless. 
Open an issue for a version 2.0

baywet: reverted, and logged as #184

src/overlay.md
baywet: updated.

mikeschinkel:
@baywet — #justfyi I am working on an updated [overlay.md](https://github.com/baywet/Overlay-Specification/blob/feat/copy-node/src/overlay.md) to illustrate what I mentioned on the call today. I am adding some examples and changing some headings for consistency because with a table of examples the inconsistency stands out more. I also changed some text from passive to active voice because the ghost of my technical writing professor in university is always sitting on my shoulder commenting on my passive voice.  

Hope to have this available very soon, though I am not 100% sure how to submit a PR to a PR. Should I submit a PR to [baywet/Overlay-Specification/feat/copy-node](https://github.com/baywet/Overlay-Specification/tree/feat/copy-node) and then mention that I did so here?

src/overlay.md
```diff
@@ -115,7 +115,8 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
+| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting locations to copy target nodes to in the target document. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. The result of the expression MUST be a object or array. |
```
ralfhandl:
```suggestion
| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
```
Remove restriction to object or array, `update` allows any type. It should be possible to copy a string from the `tags` array of one operation into the `tags` array of another operation.

mikeschinkel:
@baywet — I am getting my IDE flagging `x-oai-traits: ['paged']` as not valid and as I am not familiar with the syntax I asked ChatGPT and it said this, so you might want to look at it more closely?

> _The selector `$.paths.*.get[?@.x-oai-traits.paged]` for the `x-oai-traits` example currently defines `['paged']` (array), but the JSONPath selector checks for `.paged`, implying an object. It might not match as written. You may want to verify whether the intended shape was an object (`{ paged: true }`) or adjust the JSONPath for array membership._

ralfhandl:
> I am getting my IDE flagging x-oai-traits: ['paged'] as not valid

@mikeschinkel PR #118 contains a fix for that example.

schemas/v1.1-dev/schema.yaml
ralfhandl, line 1: Wrong folder, should be `v1.1-dev`.

scripts/validate.mjs
ralfhandl, line 1: Should not be necessary

versions/1.1.0-dev.md
ralfhandl, line 1: Oops, we don’t want to rename a published spec. Maybe renaming to `1.1.0-dev.md` will help?

tests/v1.1/fail/actions-invalid-description.yaml
ralfhandl, line 1: Wrong folder

ralfhandl (changes requested):

schemas/v1.1-dev/schema.yaml
baywet: moved

scripts/validate.mjs
baywet: reverted from the rebase

tests/v1.1/fail/actions-invalid-description.yaml
baywet: moved

versions/1.1.0-dev.md
baywet: fixed

versions/1.1.0-dev.md
baywet: Update for the copy: since we've updated the linguo to say it MUST match a single element, if it doesn't match any element, it'd mean it's an error.

versions/1.1.0-dev.md
```diff
@@ -21,7 +21,7 @@ The main purpose of the Overlay Specification is to provide a way to repeatably
 
 ### Overlay
 
-An Overlay is a JSON or YAML structure containing an ordered list of [Action Objects](#overlay-actions) that are to be applied to the target document. Each [Action Object](#action-object) has a `target` property and a modifier type (`update` or `remove`). The `target` property is a [[RFC9535|JSONPath]] query expression that identifies the elements of the target document to be updated and the modifier determines the change.
+An Overlay is a JSON or YAML structure containing an ordered list of [Action Objects](#overlay-actions) that are to be applied to the target document. Each [Action Object](#action-object) has a `target` property and a modifier type (`update`, `remove` or `copy`). The `target` property is a [[RFC9535|JSONPath]] query expression that identifies the elements of the target document to be updated and the modifier determines the change.
```
```suggestion
An Overlay is a JSON or YAML structure containing an ordered list of [Action Objects](#overlay-actions) that are to be applied to the target document. Each [Action Object](#action-object) has a `target` property and a modifier type (`update`, `remove`, or `copy`). The `target` property is a [[RFC9535|JSONPath]] query expression that identifies the elements of the target document to be updated and the modifier determines the change.
```

versions/1.1.0-dev.md
```diff
@@ -115,7 +115,8 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
+| <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
```
ralfhandl, line 119: Current text sounds as if `target` is only allowed to select at most one node, for example "If the `target` selects an object node".

It also doesn't explain what happens if `target` selects a mix of object and array nodes.

This is probably not possible if an overlay is applied to a valid OpenAPI Description, but easily possible if overlays are applied to non-OADs.

ralfhandl: Opened 
* #207

baywet: thanks for creating the issue. I was about to reply that I more or less tried to align the wording with the update keyword and if we want to change that, we should do it for both, separately.

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,200 @@ actions:
...
+        200:
+          description: OK
+```
+
+###### Overlay
+
+```yaml
+overlay: 1.1.0
+info:
+  title: Demonstrates variations of "copy" uses
```
lornajane: Can we have specific titles for each overlay? (sorry, it's a small thing, but they're good and people will definitely use these)

```suggestion
  title: Copy contents of an existing path to a new location
```

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,200 @@ actions:
...
```
lornajane:
```suggestion
  title: Create a path and copy the contents of an existing path to the new path
```

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,200 @@ actions:
...
```
lornajane:
```suggestion
  title: Update the path for an API endpoint
```

versions/1.1.0-dev.md
```diff
@@ -265,6 +268,200 @@ actions:
...
+##### Simple copy
+
+This example shows how to copy all operations from the `items` path item to the `some-items` path item.
+
+###### Source description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
lornajane: We use this title on all the example OpenAPI snippets, but I'm not sure what it's trying to describe. Maybe just "Example API" would make more sense here?

ralfhandl: Seems it was copied from the Traits example. There we should keep it.

```suggestion
  title: Example API
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
...
+    copy: '$.paths["/items"]'
+    description: 'copies recursively all elements from the "items" path item to the new "some-items" path item without ensuring the node exists before the copy'
+```
+
+###### Result description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
```suggestion
  title: Example API
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
...
+##### Ensure the target exists and copy
+
+This example shows how to copy all operations from the `items` path item to the `other-items` path item after first ensuring the target exists with an update action.
+
+###### Source description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
```suggestion
  title: Example API
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
...
+    copy: '$.paths["/items"]'
+    description: 'copies recursively all elements from the "items" path item to the new "other-items" path item while ensuring the node exists before the copy'
+```
+
+###### Result description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
```suggestion
  title: Example API
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
...
+1. Use an `update` action to ensure the target path item exists.
+2. Use a `copy` action to copy the source path item to the target.
+3. Use a `remove` action to delete the original source path item.
+
+###### Source description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
```suggestion
  title: Example API
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
...
+    remove: true
+    description: 'moves (renames) the "items" path item to "new-items"'
+```
+
+###### Result description
+
+```yaml
+openapi: 3.1.0
+info:
+  title: API with a paged collection
```
```suggestion
  title: Example API
```

ralfhandl (changes requested):

ralfhandl (dismissed):

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
 
 This approach allows inversion of control as to where the Overlay updates apply to the target document itself.
 
+#### Copy example
+
+Copy actions behave similarly to `update` actions but source the node to from the document being transformed. Copy `actions` MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
```
mikekistler: Something whacky with this wording: "source the node to from the document". I'd suggest a fix but I'm not really sure what this should say.

ralfhandl:
```suggestion
Copy actions behave similarly to `update` actions but source the new values to merge with the target from the document being transformed instead of providing them in the Overlay document. Copy `actions` MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
```

versions/1.1.0-dev.md
```diff
@@ -267,6 +270,200 @@ actions:
 
 This approach allows inversion of control as to where the Overlay updates apply to the target document itself.
 
+#### Copy example
+
+Copy actions behave similarly to `update` actions but source the new values to merge with the target from the document being transformed instead of providing them in the Overlay document. Copy `actions` MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
```
```suggestion
Copy actions behave similarly to `update` actions but source the new values to merge with the target from the document being transformed instead of providing them in the Overlay document. Copy actions MAY be used in sequence with `update` or `remove` actions to perform more advanced transformations like moving or renaming nodes.
```

PR #152, opened 2025-10-16
ci: adds a dependabot configuration to maintain workflow and test projects up to date

baywet:


.github/dependabot.yml
```diff
@@ -0,0 +1,16 @@
...
+    open-pull-requests-limit: 10
+  - package-ecosystem: npm
+    directory: "/"
+    schedule:
+      interval: daily
+    open-pull-requests-limit: 10
+    groups:
+      vitest:
+        patterns:
+          - "*vitest*"
```
ralfhandl, line 16: Why this group? That is not present in the OpenAPI-Specification repo's config.

baywet: this is because vitest has multiple dependencies that ship with every update. This will reduce the number of pull request we get.
We can either:
- make the same update to the main repo
- remove this grouping

I'm not attached to it :)

PR #171, opened 2025-10-20
Update workflow to run every Tuesday but only create a meeting every other week

lornajane:
The current cron runs only on the 1, 15 and 29 of the month (every 14 days) if they are Tuesdays, so it hasn't been creating meetings. This change adjusts the workflow to run every week but only create next week's meeting every OTHER week.

### DIFF
--- a/.github/workflows/agenda.yaml
+++ b/.github/workflows/agenda.yaml
@@ -3,7 +3,7 @@ name: Create meeting template
 on:
   workflow_dispatch: {}
   schedule:
-    # every two weeks on tuesday at 10AM PST (with DST)
+    # every week on tuesday at 10AM PST (with DST)
     - cron: '0 17 * * 2'
 
 jobs:


ralfhandl (approved):
Looks convincing

ralfhandl:
> @baywet I don't seem to be able to tag you as a reviewer (not sure why) but could you take a look?

@lornajane He is not a maintainer or otherwise a member of the repo. Should I add him, and with which role?

.github/workflows/agenda.yaml
```diff
@@ -4,7 +4,7 @@ on:
   workflow_dispatch: {}
   schedule:
     # every two weeks on tuesday at 10AM PST (with DST)
```
baywet:
```suggestion
    # every week on tuesday at 10AM PST (with DST)
```

PR #175, opened 2025-10-21
main: Run respec workflow only if spec versions change

ralfhandl:
To avoid noise in the source repo for spec.openapis.org

### DIFF
--- a/.github/workflows/schema-publish.yaml
+++ b/.github/workflows/schema-publish.yaml
@@ -20,7 +20,7 @@ jobs:
     steps:
       - name: Generate access token
         id: generate-token
-        uses: actions/create-github-app-token@v1
+        uses: actions/create-github-app-token@v2
         with:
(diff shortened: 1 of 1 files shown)


.github/workflows/respec.yaml
```diff
@@ -4,68 +4,70 @@ name: respec
...
+# This workflow updates the respec 'pretty' rendered versions of the spec
 # on the gh-pages branch when the corresponding markdown files change.
 #
 
-# run this on push to main
+# run this on push of new spec versions to main
 on:
   push:
+    paths:
+      - "versions/**"
```
baywet, line 15:
```suggestion
      - "versions/**"
      - "src/*"
```

(and maybe include the dev branches so it runs their as well?)

ralfhandl: This workflow creates a PR in the repo backing https://spec.openapis.org, where released spec versions are published.

We don't want that for work-in-progress specs, at least until now.

PR #176, opened 2025-10-21
v1.1-dev: sync from main

baywet:
syncs all the latest dependabot PRs to 1.1-dev


baywet:
I suggest editing your initial message there to add `closes #XXX` syntax. This way when it merges it'll take care of that automatically. Same for #157

PR #177, opened 2025-10-21
ci: adds an auto-merge workflow for dependenabot minor and patch pull request

baywet:
This adds a workflow that will set auto-merge for dependabot PRs that are semver patch or minors.
It will still require a review from a human being before it can merge, but save one click to that human being.
We'll need to enable auto-merge for this workflow to work properly.

I've used this workflow across dozens of repositories over the years, we could imaging hoisting it to the .github repo so it's available on all repositories for the org

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,7 +21,7 @@
         "markdownlint-cli": "^0.45.0",
         "mdv": "^1.3.4",
         "prettier": "^3.6.2",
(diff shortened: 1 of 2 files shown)


ralfhandl (approved):

baywet:
Here is a recent example of the result https://github.com/microsoft/kiota-java/pull/2045

PR #180, opened 2025-10-21
ci: sets up testing infra for 1.1 version

baywet:
this pull request sets up the testing infrastructure for v1.1.
It also adds the schema for v1.1, only notable difference is the improvement usage of oneOf since properties shouldn't be used together.
It also corrects some of the test cases in 1.0 to add a NO-OP update, so they are sematically meaningfull and ONLY test the aspect they should be testing.
All those changes have been abstracted away from #150, which will need to be rebased after this gets merged


tests/v1.0/pass/actions-description.yaml
```diff
@@ -5,3 +5,4 @@ info:
 actions:
   - target: '$' # Root of document
     description: this is an action description
+    update: {}
```
ralfhandl: We shouldn't change v1.0 test cases in the `v1.1-dev` branch. 

It is odd to have the v1.0 schema and test cases at all in this branch: we should discuss this aspect of our branch strategy. In the OpenAPI-Specification and Arazzo-Specification repos the schemas-under-development live in the `src` folder, and the schema tests only test the schemas-under-development for the spec version developed in that branch.

baywet: Reverted, again this was due to the iterative nature of the work, and I wanted to ensure that we thought this was undesired before reverting it so it does not get lost in the PR gymnastics inadvertently :)

ralfhandl (approved):

PR #181, opened 2025-10-21
main: use schema test coverage provided by @hyperjump/json-schema-coverage

ralfhandl:
- Align schema tests with https://github.com/OAI/OpenAPI-Specification/blob/dev/tests/schema/schema.test.mjs (first commit)
- Refactor to remove duplicate code (second commit)

### DIFF
--- a/.github/workflows/schema-tests.yaml
+++ b/.github/workflows/schema-tests.yaml
@@ -24,7 +24,7 @@ jobs:
 
       - uses: actions/setup-node@v6 # setup Node.js
         with:
-          node-version: "22.x"
+          node-version: "20.x"
 
       - name: Install dependencies
         run: npm ci


.github/workflows/schema-tests.yaml
```diff
@@ -24,12 +24,12 @@ jobs:
 
       - uses: actions/setup-node@v6 # setup Node.js
         with:
-          node-version: "20.x"
+          node-version: "22.x"
```
baywet: there are other places where we are using version 20 in this repo, we should probably align that if we're to change the version in this PR.

ralfhandl: Should I revert this here so you can change the Node version across all scripts in one go?

baywet: let's reduce the ceremony here if you don't mind. I'd be happy if you changed all versions as part of this PR as well.

ralfhandl: No, I don't want to touch stuff in this PR that is unrelated to code coverage: one PR, one topic

baywet: ah, just saw your commit, alright then, I'll submit another PR.

PR #198, opened 2025-10-28
Workflow to complain if a pull request changes restricted files

ralfhandl:
Not a required status check yet, just a traffic light.

### DIFF
--- a/.github/workflows/check-restricted-files.yaml
+++ b/.github/workflows/check-restricted-files.yaml
@@ -8,7 +8,6 @@ name: check-restricted-files
 on:
   pull_request:
     paths:
-      - "schemas/v[0-9].[0-9]/*.yaml"
       - "versions/[0-9].[0-9].[0-9].md"
 
 jobs:
@@ -34,6 +33,6 @@ jobs:
 
           echo This PR contains changes to files that should not be changed:
           echo
-          git diff --compact-summary origin/${{ github.event.pull_request.base.ref }} origin/${{ github.event.pull_request.head.ref }} -- schemas/v[0-9].[0-9]/*.yaml versions/[0-9].[0-9].[0-9].md
+          git diff --compact-summary origin/${{ github.event.pull_request.base.ref }} origin/${{ github.event.pull_request.head.ref }} -- versions/[0-9].[0-9].[0-9].md
 
           exit 1


.github/workflows/check-restricted-files.yaml
```diff
@@ -0,0 +1,39 @@
...
+    steps:
+      - uses: actions/checkout@v5 # checkout repo content
+        with:
+          fetch-depth: 0
+
+      - name: Check changed files
+        shell: bash
+        run: |
+          if [[ "${{ github.event.pull_request.head.repo.full_name }}" == "OAI/Overlay-Specification" ]] && \
+             [[ "${{ github.event.pull_request.base.repo.full_name }}" == "OAI/Overlay-Specification" ]]; then
```
baywet, line 25: I'd move those conditions (repo check) to the workflow definition itself, it'd make things more readable.

You might be able to more the ref validation as well, I'm not sure whether the regex will be translatable though.

ralfhandl: > I'd move those conditions (repo check) to the workflow definition itself, it'd make things more readable.

How?

.github/workflows/check-restricted-files.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+on:
+  pull_request:
+    paths:
+      - "schemas/v[0-9].[0-9]/*.yaml"
+      - "versions/[0-9].[0-9].[0-9].md"
+
+jobs:
+  check-files:
+    runs-on: ubuntu-latest
```
baywet, line 15:
```suggestion
    if: github.repository == 'OAI/Overlay-Specification'
    runs-on: ubuntu-latest
```

.github/workflows/check-restricted-files.yaml
baywet: https://github.com/OAI/Overlay-Specification/pull/198/files#r2469512626

.github/workflows/check-restricted-files.yaml
ralfhandl: That would only get rid of the base repo check, not the head repo check.

And having two kinds of checks in two different places does not make things more readable for me.

baywet: I just meant that as an example as the question was not super specific.

```suggestion
    if: github.repository == 'OAI/Overlay-Specification' && github.event.pull_request.head.repo.full_name == 'OAI/Overlay-Specification'
    runs-on: ubuntu-latest
```

ralfhandl: This would make the workflow _not_ run if someone creates a PR in a fork to change a published specification.

We want the workflow to run in this case and _fail_.

ralfhandl: I actually started with the same idea as you and then figured out that it would not work 🤦🏻‍♂️

ralfhandl: Point is:
- Workflow has to run on all PRs changing a restricted file
- Changing a restricted file is OK if and only if the change is done in a release branch within this repo, then it exits with code 0.

baywet: ah because you want to exit early only when the whole set of conditions is met.

what if we had something like this then

```suggestion
    runs-on: ubuntu-latest
    if: github.repository != 'OAI/Overlay-Specification' ||
        github.event.pull_request.head.repo.full_name != 'OAI/Overlay-Specification' ||
        github.event.pull_request.base.ref != 'main' ||
        !(github.event.pull_request.head.ref =~ '^v[0-9]+\.[0-9]+\.[0-9]+-rel$')
```

ralfhandl: Apparently there is no `=~` operator in workflow expressions:
* https://docs.github.com/en/actions/reference/workflows-and-actions/expressions#operators

baywet: I've spent the last 15 minutes search for a reliable reference for that... I'm pretty sure at this point that GitHub copilot just hallucinated on me... including some made up "reference links" 🤦

Since there doesn't seem to be a way to express the whole condition, we'd have fall through cases. Which means the condition cannot be hoisted out of the script. Thanks for iterating with me on this one. I don't have any further suggestions.

PR #215, opened 2025-10-31
fix/ref rfc 9535

### LINKED ISSUE #88: Wildcard JSONPath example in spec text seems incorrect
I was trying to implement the spec as part of Apigee's [apigee-go-gen](https://github.com/apigee/apigee-go-gen) tool, and ran into this issue while creating some test cases.

The example JSON Path expression from this example: https://github.com/OAI/Overlay-Specification/blob/main/versions/1.0.0.md?plain=1#L197 could not be parsed successfully by library at from: https://github.com/vmware-labs/yaml-jsonpath

`$.paths.*.get.parameters[?@.name=='filter' && @.in=='query']`

I had to modify it by adding grouping parenthesis, and then it worked:

`$.paths.*.get.parameters[?(@.name=='filter' && @.in=='query)']`

I also tried with various other online tools and it seems like they all need parenthesis. I am not a JSONPath expert, so I am opening this issue. Please clarify or fix the example.

As a separate nitpick ... if I understand correctly, the example is attempting to change the schema property to use a $ref instead of it being inlined. However, that's not what's it's actually doing, it's adding the $ref field to the schema object. So you end up with a schema object that is both in-lined, and also $ref. Should not the example first remove the schema object, and then update the operation with the schema using $ref, ... e.g.

```
overlay: 1.0.0
info:
  title: Update many objects at once
  version: 1.0.0
actions:
  - target: $.paths.*.get
    update:
      x-safe: true
  - target: $.components.schemas
    update:
      filterSchema:
        type: string
        default: available
        enum:
          - available
          - pending
          - sold
  - target: $.paths.*.get.parameters[?(@.name=='filter' && @.in=='query')].schema
    remove: true
  - target: $.paths.*.get.parameters[?(@.name=='filter' && @.in=='query')]
    update:
      schema:
        $ref: '#/components/schemas/filterSchema'
```

baywet:
fixes #88

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ If you are looking for tools to use with Overlays, try these:
 
 ## Support for RFC9535 JSONPath
 
-[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC).
+[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), implement additional capabilities (superset of the RFC), or support only a subset of the RFC.
 
 In case the tool being use is not fully compliant with RFC9535, users MIGHT have to update some JSONPath query expressions to accommodate for the implementation in use.
 


versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
 Overlay files MAY choose to follow the convention of a `purpose.overlay.yaml` file naming pattern.
 Other file naming conventions are also supported.
 
+### RFC9535 compliance
+
+[[RFC9535|RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC). A tool or library MUST fully implement [[RFC9535|RFC9535]] when parsing and expanding JSONPath query expressions to be compliant with the Overlay specification.
+
+Users SHALL refrain from using additional capabilities offered by the tool's JSONPath implementation to avoid any interoperability issues with their Overlay Documents.
+
+In case the tool being use is not fully compliant with RFC9535, users MIGHT have to update some JSONPath query expressions to accommodate for the implementation in use.
```
mikekistler: I'm not sure this statement is needed / appropriate. When a tool doesn't comply with this spec, I think it would be better to recommend that users find a compliant tool than adjust their usage to fit the non-compliant tool.

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
...
+```jsonpath
+$.paths.*.get.parameters[?@.name=='filter' && @.in=='query']
+```
+
+Might require additional optional parenthesis with some implementations like so:
+
+```jsonpath
+$.paths.*.get.parameters[?(@.name=='filter' && @.in=='query)']
+```
+
```
mikekistler: This too seems inappropriate for the spec, though I agree it is useful information. Could we put this in the README instead, as a subsection of "Tools that Support Overlays"?

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
 Overlay files MAY choose to follow the convention of a `purpose.overlay.yaml` file naming pattern.
 Other file naming conventions are also supported.
 
+### RFC9535 compliance
+
+[[RFC9535|RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC). A tool or library MUST fully implement [[RFC9535|RFC9535]] when parsing and expanding JSONPath query expressions to be compliant with the Overlay specification.
```
```suggestion
[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC). A tool or library MUST fully implement [[RFC9535]] when parsing and expanding JSONPath query expressions to be compliant with the Overlay specification.
```
No need for defining an "alias" text that is identical to the default text.

versions/1.1.0-dev.md
baywet: I also considered putting this as an appendix in the beginning of writing this. But since I have parts that are normative above, I put it in the full specification. Maybe a better approach would be to keep the informational parts and the example as an appendix, and get rid of the normative parts of this section?

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
...
+
+Users SHALL refrain from using additional capabilities offered by the tool's JSONPath implementation to avoid any interoperability issues with their Overlay Documents.
```
```suggestion
Interoperable Overlay Documents MUST use RFC9535 JSONPath query expressions and MUST NOT use tool-specific JSONPath extensions.
```

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
...
+
+In case the tool being use is not fully compliant with RFC9535, users MIGHT have to update some JSONPath query expressions to accommodate for the implementation in use.
+
+### Example of updates to JSONPath
```
ralfhandl: I would call this a restriction and not an "update" 😎 

This whole section should rather go into the README, as proposed by @mikekistler.

ralfhandl (commented):
+1 with a few minor suggestions

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
...
 
+### RFC9535 compliance
+
+[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC). A tool or library MUST fully implement [[RFC9535]] when parsing and expanding JSONPath query expressions to be compliant with the Overlay specification.
+
+Interoperable Overlay Documents MUST use RFC9535 JSONPath query expressions and MUST NOT use tool-specific JSONPath extensions.
+
+In case the tool being use is not fully compliant with RFC9535, users MIGHT have to update some JSONPath query expressions to accommodate for the implementation in use.
+
+### Example of updates to JSONPath
```
```suggestion
#### Example of Differences between JSONPath Implementations
```

versions/1.1.0-dev.md
```diff
@@ -282,6 +282,28 @@ The extensions may or may not be supported by the available tooling, but those m
...
+
+This example JSONPath query expression:
+
+```jsonpath
+$.paths.*.get.parameters[?@.name=='filter' && @.in=='query']
+```
+
+Might require additional optional parenthesis with some implementations like so:
```
```suggestion
might require additional optional parenthesis with some implementations like so:
```

versions/1.1.0-dev.md
ralfhandl: This example subsection should move to the README.

The compliance section should be in the spec.

versions/1.1.0-dev.md
baywet: moved this paragraph and the example to the readme

versions/1.1.0-dev.md
baywet: moved this paragraph and the example to the readme

mikekistler (dismissed):
Looks good! 👍

README.md
```diff
@@ -30,6 +30,24 @@ If you are looking for tools to use with Overlays, try these:
 
 (Is something missing from the list? Send us a pull request to add it!)
 
+## Support for RFC9535 JSONPath
+
+[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), or implement additional capabilities (superset of the RFC).
```
ralfhandl:
```suggestion
[[RFC9535]] is a recent specification and libraries implementing JSONPath support might predate the RFC. Those libraries might differ entirely (expressions syntax is incompatible), implement additional capabilities (superset of the RFC), or support only a subset of the RFC.
```

PR #225, opened 2025-11-04
Describe behavior for primitive values in update and remove

### LINKED ISSUE #57: Allow updating and removing primitive values
The current text does not allow directly targeting a primitive value, for example a certain string in the `tags` array of an Operation Object, and thus does not allow a convenient/workable way to remove or replace a certain tag name without knowing all other tag names in the `tags` field.

### Proposal

Allow `target` JSON Path expressions to select nodes of all JSON types.

For example allow

```yaml
overlay: 1.0.0
info:
  title: Remove a string from the tags array of all operations
  version: 1.0.0
actions:
  - target: $.paths.*.*.tags[?@ == 'dummy']
    remove: true
```

and

```yaml
overlay: 1.0.0
info:
  title: Replace a string in the tags array of all operations
  version: 1.0.0
actions:
  - target: $.paths.*.*.tags[?@ == 'Foo']
    update: Bar
```

### LINKED ISSUE #70: `update` should not be `type: Any` based on description
https://github.com/OAI/Overlay-Specification/blob/d029f0d7d2c8861e00de4467ab795e86d9e35824/versions/1.0.0.md?plain=1#L125

This text indicates the `type` of `update` should only be `object | array` and _not_ `Any`

ralfhandl:
This applies the rules for merging property values to non-object `update` values and extends `remove` to primitive target nodes.

Note: these features are already fully supported by
- [Apigee Go Gen](https://apigee.github.io/apigee-go-gen/transform/commands/oas-overlay/)
- [bump-cli](https://www.npmjs.com/package/bump-cli#the-overlay-command)
- [BinkyLabs.OpenApi.Overlays - dotnet](https://github.com/BinkyLabs/openapi-overlays-dotnet)
- [speakeasy](https://www.speakeasy.com/docs/speakeasy-reference/cli/overlay)

and mostly supported by
- [OAS Patcher](https://github.com/mcroissant/oas_patcher)
- [openapi-format](https://github.com/thim81/openapi-format)

See https://github.com/ralfhandl/overlay-experiments for tests of these features with the mentioned Overlay tools.

Fixes #57
Closes #70

### DIFF
--- a/versions/1.1.0-dev.md
+++ b/versions/1.1.0-dev.md
@@ -169,17 +169,15 @@ actions:
 Alternatively, where only a small number of updates need to be applied to a large document, each [Action Object](#action-object) MAY be more targeted.
 
 ```yaml
-overlay: 1.0.0
+overlay: 1.1.0
 info:
   title: Targeted Overlay
   version: 1.0.0
 actions:
-  - target: $.paths['/foo'].get
-    update:
-      description: This is the new description
-  - target: $.paths['/bar'].get
-    update:
-      description: This is the updated description
+  - target: $.paths['/foo'].get.description
+    update: This is the new description
+  - target: $.paths['/bar'].get.description
+    update: This is the updated description
   - target: $.paths['/bar']
     update:
       post:


versions/1.1.0-dev.md
```diff
@@ -115,17 +115,13 @@ This object represents one or more changes to be applied to the target document
...
+| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected object. If the `target` selects an array node, the value of this field MUST be an array to concatenate with the selected array, or an object or primitive value to append to the selected array. If the `target` selects a primitive node, the value of this field MUST be a primitive value to replace the selected node. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
 | <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
```
baywet: This probably needs the same changes as the update one

baywet (dismissed):
Than you for making the changes

baywet (dismissed):

versions/1.1.0-dev.md
```diff
@@ -115,19 +115,14 @@ This object represents one or more changes to be applied to the target document
...
-| <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that each of the target objects or arrays MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-update"></a>update | Any | If the `target` selects object nodes, the value of this field MUST be an object with the properties and values to merge with each selected object. If the `target` selects array nodes, the value of this field MUST be an array to concatenate with each selected array, or an object or primitive value to append to each selected array. If the `target` selects primitive nodes, the value of this field MUST be a primitive value to replace each selected node. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
```
mikekistler, line 118: It's so hard to read these diffs. But it looks like this is changing more than just the behavior for primitive nodes. I _think_ the old version says:
> If the `target` selects array nodes, the value of this field MUST be an entry to append to each selected array.

and the new version says:
> If the `target` selects array nodes, the value of this field MUST be an array to concatenate with each selected array

If I'm reading this correctly, now we're allowing concatenation of arrays where before we did not.

I don't have a problem with that, but if this is indeed intended then I should it should be called out explicitly in the description.

ralfhandl: We are allowing concatenation of arrays _in addition_ to appending an entry.

Concatenation of arrays already happened when merging the array values of two properties, now we are pulling that behavior up to `update` and target values without a „wrapper“ object.

Fun act: [all Overlay tools I tested](https://github.com/ralfhandl/overlay-experiments/tree/main?tab=readme-ov-file#overlay-experiments) support the so far undocumented „update array with array“, and not all support the explicitly documented „update array with object entry“ and „update array with primitive entry“:

image

PR #230, opened 2025-11-05
Zero targets, or more than one

ralfhandl:
Explicitly state that zero target nodes is a no-op.

Rephrase text to also cover two or more target nodes.

### DIFF
--- a/versions/1.1.0-dev.md
+++ b/versions/1.1.0-dev.md
@@ -107,7 +107,7 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 
 #### Action Object
 
-This object represents one or more changes to be applied to the target document at the location defined by the target JSONPath expression.
+This object represents one or more changes to be applied to the target document at the locations defined by the target JSONPath expression.
 
 ##### Fixed Fields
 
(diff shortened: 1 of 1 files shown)


versions/1.1.0-dev.md
```diff
@@ -107,19 +107,21 @@ This object MAY be extended with [Specification Extensions](#specification-exten
...
 | ---- | :----: | ---- |
 | <a name="action-target"></a>target | `string` | **REQUIRED** A RFC9535 JSONPath query expression selecting nodes in the target document. |
 | <a name="action-description"></a>description | `string` | A description of the action. [[CommonMark]] syntax MAY be used for rich text representation. |
-| <a name="action-update"></a>update | Any | If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
+| <a name="action-update"></a>update | Any | If the `target` selects object nodes, the value of this field MUST be an object with the properties and values to merge with each selected object. If the `target` selects array nodes, the value of this field MUST be an entry to append to each selected array. This field has no impact if the `remove` field of this action object is `true` or if the `copy` field contains a value. |
```
baywet, line 118: Aren't we missing the case when we have multiple entries to append to the array?

ralfhandl: That's in the other PR, #225.

versions/1.1.0-dev.md
```diff
@@ -107,19 +107,21 @@ This object MAY be extended with [Specification Extensions](#specification-exten
...
-| <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that the target object or array MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that each of the target objects or arrays MUST be removed from the the map or array it is contained in. The default value is `false`. |
 
-The result of the `target` JSONPath expression MUST be zero or more objects or arrays (not primitive types or `null` values). Should the `target` JSONPath result in selecting two or more nodes, they MUST be either all objects or all arrays.
+The result of the `target` JSONPath expression MUST be zero or more objects or arrays (not primitive types or `null` values).
+If the `target` JSONPath expression selects zero nodes, the action succeeds without changing the target document.
```
mikekistler, line 123: Does this mean that I cannot "create" new nodes in the document? Suppose my target is `$.paths./health' and the update value is a "get" operation and there is no "/health" path in the document. Does this silently ignore the operation without adding the new "get" operation?

ralfhandl: You can create new nodes in the document by adding new properties to existing objects, or by adding new entries to existing arrays, as shown in the Traits Example.

Targeting a non-existing node is a no-operation, consistently implemented by the six Overlay tools I test here: https://github.com/ralfhandl/overlay-experiments, files `examples/update-non-existing-target-*`. Each tool silently copies the input to the result without changing it.

lornajane: This makes sense to me - I would target the node that does exist that I wanted to add to.

PR #232, opened 2025-11-05
docs: changes description for x- field as its not supported for the object

baywet:
Context: I'm using the compliant sets for unit tests in here https://github.com/BinkyLabs/openapi-overlays-dotnet/pull/139

One of the examples is not compliant because the info object does NOT support a description field today.

This PR changes it for an x- so the example is fully compliant with the specification.

### DIFF
--- a/compliant-sets/replace-servers-for-sandbox/overlay.yaml
+++ b/compliant-sets/replace-servers-for-sandbox/overlay.yaml
@@ -1,7 +1,7 @@
-overlay: 1.0.0
+overlay: 1.1.0
 info:
   title: Change server URL
-  x-description: Replace servers list with a single sandbox or local development URL for the API
+  description: Replace servers list with a single sandbox or local development URL for the API
   version: 1.0.0
 actions:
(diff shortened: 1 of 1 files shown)


compliant-sets/replace-servers-for-sandbox/overlay.yaml
```diff
@@ -1,7 +1,7 @@
 overlay: 1.0.0
```
baywet:
```suggestion
overlay: 1.1.0
```

compliant-sets/replace-servers-for-sandbox/overlay.yaml
```diff
@@ -1,7 +1,7 @@
...
 info:
   title: Change server URL
-  description: Replace servers list with a single sandbox or local development URL for the API
+  x-description: Replace servers list with a single sandbox or local development URL for the API
```
baywet:
```suggestion
  description: Replace servers list with a single sandbox or local development URL for the API
```

PR #233, opened 2025-11-05
feat: adds description field to overlay info object

baywet:
This pull request adds a description field to the info object.
While working on implementing additional tests in clio, I discovered that one of the compliant tests was actually non compliant with the specification (see #232).
Having a description field in the info object aligns with the OpenAPI specification and seems to be desirable given we (spec authors) already "started using this field".

While chatting this over with @ralfhandl I originally suggested to also add a summary field to align with OpenAPI further. He was not sold on the idea. Happy either way based on the feedback here. Let me know and I'll add this other field as well to the pull request.

### DIFF
--- a/tests/v1.1-dev/pass/minimal.yaml
+++ b/tests/v1.1-dev/pass/minimal.yaml
@@ -2,5 +2,6 @@ overlay: 1.1.0
 info:
   title: Minimal
   version: 1.0.0
+  description: A minimal valid overlay specification.
 actions:
   - target: '$' # Root of document


versions/1.1.0-dev.md
```diff
@@ -102,6 +102,7 @@ The metadata MAY be used by the clients if needed.
 | ---- | :----: | ---- |
 | <a name="info-title"></a>title | `string` | **REQUIRED**. A human readable description of the purpose of the overlay. |
 | <a name="info-version"></a>version | `string` | **REQUIRED**. A version identifier for indicating changes to the Overlay document. |
+| <a name="info-description"></a>description | `string` | A description of the Overlay Document. [[CommonMark]] syntax MAY be used for rich text representation. |
```
ralfhandl, line 105: `description` should be after `title` and before `version `

baywet: why? I thought we wanted the required fields first.

ralfhandl: I wasn’t aware of that, OpenAPI does it differently, and I looked at its Info Object.

baywet: I think @lornajane mentioned that when I worked on the copy proposal if I remember correctly :)

lornajane: I didn't intend it as a rule. In the context of the action object, it made sense to have the target and description (which should always be present although actually description is technically optional) together before the update/copy/remove actions and that's the ordering I requested there.

In this context, title and description together probably does make sense and would be familiar to openapi users.

baywet:
@lornajane sorry, CLI monkey here. Updated the original comment. Let me know if you have any additional comments or questions.

PR #234, opened 2025-11-05
chore: add dark theme support

baywet:
this removes most differences with the equivalent from the main repository in an effort to restore support for dark theme.
When rendering locally, the bash script is giving me a difficult time on WSL, and the styling looks quite different.
I'm not sure whether it's expected for a local build, or if I broke something, so I'd really like another pair of eyes from somebody who has more experience with that before we merge it.

> Note: I didn't restore the "other versions" thing but we'll probably need to restore that as well before releasing 1.1

### DIFF
--- a/scripts/md2html/md2html.js
+++ b/scripts/md2html/md2html.js
@@ -93,7 +93,6 @@ function preface(title,options) {
 
     // SEO
     preface += `<meta charset="UTF-8">\n<title>${md.utils.escapeHtml(title)}</title>`;
-    preface += '<meta name="description" content="The Overlay Specification defines a document format for information that augments an existing OpenAPI description yet remains separate from the OpenAPI description\'s source document(s).">\n';
 
     // ReSpec
     preface += '<meta name="color-scheme" content="light dark">';
@@ -106,8 +105,8 @@ function preface(title,options) {
     preface += '</style>';
     preface += `<h1 id="title">${title.split('|')[0]}</h1>`;
     preface += `<p class="copyright">Copyright © ${options.publishDate.getFullYear()} the Linux Foundation</p>`;
-    preface += `<section class="notoc" id="abstract"><h2>${abstract}</h2>`;
-    preface += 'The Overlay Specification defines a document format for information that augments an existing [[OpenAPI]] description yet remains separate from the OpenAPI description’s source document(s).';
+    preface += `<section class="notoc" id="abstract"><h2>${abstract}</h2>\n`;
+    preface += options.abstract.join('\n');
     preface += '</section>';
     preface += '<section class="override" id="sotd" data-max-toc="0">';
     preface += '<h2>Status of This Document</h2>';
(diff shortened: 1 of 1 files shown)


scripts/md2html/md2html.js
```diff
@@ -93,32 +93,21 @@ function preface(title,options) {
...
-    preface += 'a.bibref { text-decoration: underline;}';
+    preface += fs.readFileSync(path.resolve(__dirname,'main.css'),'utf8').split(/\r?\n/).join(' ');
     preface += fs.readFileSync(path.resolve(__dirname,'gist.css'),'utf8').split(/\r?\n/).join(' ');
     preface += '</style>';
     preface += `<h1 id="title">${title.split('|')[0]}</h1>`;
     preface += `<p class="copyright">Copyright © ${options.publishDate.getFullYear()} the Linux Foundation</p>`;
-    preface += `<section class="notoc" id="abstract"><h2>${abstract}</h2>\n`;
-    preface += options.abstract.join('\n');
+    preface += `<section class="notoc" id="abstract"><h2>${abstract}</h2>`;
+    preface += 'The Overlay Specification defines a document format for information that augments an existing [[OpenAPI]] description yet remains separate from the OpenAPI description’s source document(s).';
```
ralfhandl: Please revert this change, we don't want to hard-code the Overlay abstract in this script, instead it is taken from the Markdown source.

scripts/md2html/md2html.js
```diff
@@ -160,7 +149,7 @@ function getPublishDate(m) {
             let v = $(c[0]).text();
             let d = $(c[1]).text();
             argv.subtitle = v;
-            if (d !== 'TBD') result = new Date(d);
+            if (d !== 'TBA') result = new Date(d);
```
ralfhandl: Please revert, I  forgot to fix this in the OpenAPI repo.

scripts/md2html/md2html.js
```diff
@@ -313,4 +289,5 @@ for (let l in lines) {
 
 s = preface(`Overlay Specification v${argv.subtitle} | Introduction, Definitions, & More`,argv)+'\n\n'+lines.join('\n');
 let out = md.render(s);
+out = out.replace(/\[([RGB])\]/g,'&#91;$1&#93;');
```
ralfhandl: Please remove, this is not necessary here. It is special fiddling needed for OpenAPI specs up to 3.0.3 and 3.1.0 for a style example 

    color[R]=100&color[G]=200&color[B]=150

scripts/md2html/md2html.js
```diff
@@ -181,24 +170,11 @@ let inTOC = false;
...
-    if (line.startsWith('## Abstract')) { 
-        inAbstract = true;
-        line = '';
-    }
-    else if (line.startsWith('#')) inAbstract = false; 
-    else if (inAbstract) { 
-        argv.abstract.push(line);
-        line = '';
-    }
-
```
ralfhandl, line 201: Please revert, this extracts the "Abstract" section from the Markdown source and uses it instead of a hard-coded abstract.

The OpenAPI specs need the hard-coding because the old Markdown sources don't contain individual abstracts.

PR #237, opened 2025-11-07
Describe steps for publishing a release and starting a new release

ralfhandl:
- [x] Describe steps in CONTRIBUTING.md
- [x] Add script `adjust-release-branch.sh`
- [x] Add script `start-release.sh`
- [x] Remove obsolete script `format-markdown.sh`
- [x] Use `markdownlint-cli2 --fix` in npm script `format-markdown`, do not fix published spec versions

### DIFF
--- a/scripts/start-release.sh
+++ b/scripts/start-release.sh
@@ -1,0 +1,98 @@
+#!/usr/bin/env bash
+
+# Author: @ralfhandl
+
+# Run this script from the root of the repo. It is designed to be run manually in the main branch.
+
+if [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+  echo "Usage: $0 <nextVersion>"
+  echo "Example: $0 1.1.1"
+  exit 1
+fi
+
+nextVersion=$1
+minor=$(echo $nextVersion | cut -d. -f1,2)
+nextPatch=$(echo $nextVersion | cut -d. -f3)
+
+# Find last published spec version for this minor version
+lastSpec=$(ls -1 versions/*.md | grep -E "/$minor\.[0-9].md" | tail -1)
+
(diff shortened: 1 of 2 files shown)


scripts/adjust-release-branch.sh
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+
+# Author: @ralfhandl
+
+# Run this script from the root of the repo. It is designed to be run manually in a release branch.
+
+branch=$(git branch --show-current)
+today=$(date +%Y-%m-%d)
+
+if [[ ! $branch =~ ^v[0-9]+\.[0-9]+\.[0-9]+-rel$ ]]; then
```
baywet, line 10: nit: why not use `\d` instead of  ranges?

scripts/start-release.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+
+# Author: @ralfhandl
+
+# Run this script from the root of the repo. It is designed to be run manually in the main branch.
+
+if [[ ! $1 =~ ^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
```
baywet, line 7: nit: same comment about `\d`

scripts/adjust-release-branch.sh
ralfhandl: Bash uses [POSIX extended regular expressions](https://www.gnu.org/software/findutils/manual/html_node/find_html/posix_002degrep-regular-expression-syntax.html) which do not have `\d`, instead they have `[[:digit:]]` which is more verbose and less expressive than `[0-9]`.

baywet: Til that POSIX regex do not support \d....

PR #238, opened 2025-11-07
feat: adds a parameter object for the action object

### LINKED ISSUE #33: Environment Variables in definition

### LINKED ISSUE #136: String Interpolation for Update Values
Was on call with @lornajane today and we discussed [Environment Variables](https://github.com/OAI/Overlay-Specification/issues/33) and the topic of string interpolation came up.  

Lorna [proposed](https://github.com/OAI/Overlay-Specification/issues/33#issuecomment-2936023796) using a `$env.VAR_NAME` syntax, and while we were looking around for prior art we found what appears to be an example of [string interpolation in the Arazzo examples](https://spec.openapis.org/arazzo/v1.0.1.html#examples) which I will copy here, specifically `"couponCode": "{$inputs.coupon_code}",`:

**Example from Arazzo:**
```
  contentType: application/json
  payload: |
    {
      "petOrder": {
        "petId": "{$inputs.pet_id}",
        "couponCode": "{$inputs.coupon_code}",
        "quantity": "{$inputs.quantity}",
        "status": "placed",
        "complete": false
      }
    }
```
The above example does not have any suffix or prefix literals, but I can imagine that if could:

**Example modified to show suffixes and prefixes:**
```
  contentType: application/json
  payload: |
    {
      "petOrder": {
        "petId": "{$inputs.pet_id}000",
        "couponCode": "coupon-{$inputs.coupon_code}",
        "quantity": "{$inputs.quantity}",
        "status": "placed",
        "complete": false
      }
    }
```

We both agreed there was value to incorporating string interpolation although Lorna argued against doing it sooner than later _(I will let her elaborate as to why.)_

I argued if we were to adopt Arazzo's style of variables (which I support) — e.g. `$env.VAR_NAME` — then also supporting string interpolation would be very small incremental lift for implementors with an outsized increase in utility to support more of the 80 percentile of use-cases. 

This of course assumes we do not envision that supporting it would block some future capability we plan for in the future. 

Lorna asked me to create an issue so that others could weigh in on the idea, and here it is.

baywet:
Introducing the Overlay Parameters proposal

fixes #33 
fixes #136

# Summary

This pull-request introduces new parameters for the action object. The values of those parameters can be read from the environment variables, or when missing from the default values field. It also introduces a [string interpolation](https://en.wikipedia.org/wiki/String_interpolation) syntax that needs to avoid any conflicts (see the issues linked above) so that values can be inserted anywhere in the action.

# Vision

This proposal has multiple goals:
1.	Make OpenAPI descriptions more dynamic via environment variables. This has been a long-requested ability.
2.	Make Overlay actions more scalable through reducing repetitive actions.
3.	Leverage Overlay documents to avoid adding complexity to OpenAPI and to be consistent with the ecosystem.
4.	Avoid the use of pre-processing scripts/tooling as it’s highly inefficient for everyone using overlays to have to roll out their own scripts.
5.	Allow for better interoperability between implementers through standardization (instead of custom scripts or custom extensions).

The short-term use of this would be to enable goals 1 & 2. In the long term there might be an additional opportunity to come up with “libraries of overlays” which would distribute nice additions for OpenAPI descriptions (standard paging patterns, etc…) but I’m not trying to solve for that.

# Execution

### Matrix

See this comment https://github.com/OAI/Overlay-Specification/pull/238#issuecomment-3512628113 or that one https://github.com/OAI/Overlay-Specification/issues/33#issuecomment-3487948714 

### Environment variables

See the initial issue, which has tons of great examples. I think a key recap could be “I want to use my OpenAPI description as a template for multiple hosts deployments, and I don’t know the hosts names in advances” (think about cloud deployments with tenants, etc…). This also helps enable a set of CI/CD scenarios (like deploying a temporary API on a host to run some tests, and deploying the updated documentation for the API alongside)

## 1.1 Release

This should not hold the 1.1 release. I’m happy to punt it to a 1.2 release if we think it’s going to get the 1.1 release out of the door faster. And at the same time, I’d like to continue making progress on this by getting actionable feedback from the reviewers.

## Implementation

It might not be clearly called out in the current proposal, but the intended implementation guideline is the following in order:

1. Parse the Overlay document (implementers already have most of that)
2. Foreach action that uses parameters

   1. Resolve the values of each parameter
   2. Project the values of each parameter into a “projected action” (one of the parameter values combinations + the action)
   4. Foreach “projected action” execute the string interpolation of the parameter values
   5. Insert all the projected actions in place of the original parameterized action

3. Apply all the actions, projected or regular. (implementers already have that)

This is important to enable separation of concerns, get the matrix projection right, and enable maximum reusability with the code that’s already in place for v1.0. I’m happy to rework the wording to make this clear in the specification.

# Remaining questions

## Do we have the right interpolation syntax?

I’m happy with any syntax really, if it meets the constraints we have. I’ve forked the discussion here https://github.com/OAI/Overlay-Specification/issues/136#issuecomment-3548973963 

## Should we always read Environment variables?

Since it’s in the same “parametrization domain”, I thought it’d be nice to bundle changes together. In the initial draft of this proposal, we had the ability to have constants/avoid reading the environment variables, which we’ve lost with the latest revisions. I think we should restore that ability somehow. What about a field “ignoreEnvironmentVariable” (Boolean, default false)? When set to true, the implementer would only read the defaultValues.

## Do we have the right matrix/projection syntax?

I am especially interested in the projection, matrix, and scaling aspects, as my team has been dealing with these challenges and is eager to find a solution. But it also adds great value. And users of the specification are not required to use it as it is defined.
I know this is a contentious point, and there have been suggestions to cut that aspect from the proposal.  I will not put efforts behind a version of that proposal without the matrix aspect in one form or another. I’m happy to take actionable suggestions to make it better/safer/simpler/…

## Do we need parameters references?

Ralf made a pertinent comment about the fact that if multiple actions had the same parameter name, they are effectively the same parameter since they’d read from the same environment variable. And he suggested hoisting them at the document level instead of the action. While this would reduce the repetitive definitions further, we’d still need to flag any given action is using parameters, so implementers can do the pre-processing. This would require defining parameter references at the action level.
I think Lorna made a pertinent comment in another context: we want to keep the complexity low to bolster adoption. I think the “global parameters” comment goes against that, with minimal gains. And this is something we can review in v2 if people are tired of re-defining the same parameters.

I hope this rewrite of the initial comment brings a lot of clarity, I spent a lot of time on it :)


ralfhandl:
I like the idea of adding a default value if the source is environment. I'd probably name it default or defaultValue and not just value.

versions/1.1.0-dev.md
```diff
@@ -118,6 +118,7 @@ This object represents one or more changes to be applied to the target document
...
 | <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects an object node, the value of this field MUST be an object with the properties and values to merge with the selected node. If the `target` selects an array, the value of this field MUST be an entry to append to the array. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that each of the target objects or arrays MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-parameters"></a> | [[Parameter Object](#parameter-object)][] | An array of parameter objects to use and expand the action definition. MAY contain zero or more parameters. Parameters MUST have unique names across the collection. Default `undefined`. |
```
ralfhandl: Why local to an action and not global to allow use across multiple related actions, for example when applying traits?

versions/1.1.0-dev.md
```diff
@@ -141,6 +142,25 @@ The properties of the resolved `copy` object MUST be compatible with the target
...
+
+The object represents parameters that can be used with actions.
+
+#### Fixed Fields
+
+| Field Name | Type | Description |
+| ---- | :----: | ---- |
+| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique for the action the parameter is defined. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
+| <a name="parameter-source"></a>source | `inline` or `environment` | **REQUIRED** Defines where the values for the parameters should be retrieved from. When `inline` is set, values are retrieved from the `values` property. When `environment` is set, the values are retrieved by the tool from environment variables. |
+| <a name="parameter-values"></a>values | `string[]` | **REQUIRED** when `source` is set to `inline`. MAY contain zero or more entries. When values contains multiple entries, the action is effectively projected for each of the entries of the parameter. If multiple parameters are defined for any given action, parameters effectively act as a matrix definition and the action should be projected across each dimension. |
```
ralfhandl: I have one concern with this matrix feature: it allows an "exponential explosion attack" on Overlay processors. Providing $n$ two-valued parameters in an action will result in $2^n$ executions of this action.

versions/1.1.0-dev.md
```diff
@@ -141,6 +142,25 @@ The properties of the resolved `copy` object MUST be compatible with the target
...
+| <a name="parameter-separator"></a>separator | `string` | Only applicable when `source` is set to `environment`. Defaults `undefined` (no splitting is applied). When defined, the separator is used to split the environment variable value and extra multiple values. |
+
+#### Parameter Expressions
+
+Parameter expressions MAY be used in any field defined by the Action Object except for parameters. The syntax for parameter expressions is the following `${parameter_name}`. ONLY parameter expressions with a matching parameter definition for the action MAY be replaced by their values.
```
ralfhandl: What about structured parameters, for example when I need one value to parameterize the target and other values to parameterize the update object? For single-valued parameters I can just invent a naming convention, but for multi-valued parameters that would result in a matrix and not a linear correlation.

So if we want multi-values parameters, allow them to have object values and allow dot notation.

ralfhandl: Why the dollar before the curlies? OpenAPI path template parameters and server variables only use curlies around the parameter name.

versions/1.1.0-dev.md
mikekistler: I like @ralfhandl 's suggestion to make these global rather than within the action.

baywet: The reason why I didn't make the parameters global is because having the parameters on the action makes it easy for any implementer to detect parameters are used for any given action and expansion is required.

If we want to make global parameters a thing, I think we then need to introduce a parameter reference mechanism. Where we'd mandate that the action has a reference entry for a global parameter to be used and for the expansion + interpolation to happen.

Thoughts?

versions/1.1.0-dev.md
baywet: I didn't want to have a complex interpolation syntax so it remained easy for implementers to support it properly.

Likewise, I originally limited the values to strings to avoid complex serialization mechanics when sourcing from environment variables.

I think this proposal remains functional and flexible enough that we could ship it in a specific minor version of the standard and then expand upon it if we get demand for that.

As for making the whole update value parameterized as an object value, I believe in that case people would be better off with multiple actions.

baywet: I'm not married to this specific syntax. I want to maintain a syntax that's simple enough it can be implemented via string replace, not even regex if we can.

Also that syntax needs to be different ythe path template sin OAI so implementers of overlay parameters can tell the difference in that specific scenario.

versions/1.1.0-dev.md
baywet: This would only be an issue with environment variables when using the separator.

And by definition environment variables are defined locally, so I'm not sure if I follow the attack vector here?

baywet:
> I like the idea of adding a default value if the source is environment. I'd probably name it default or defaultValue and not just value. 

Differentiating the field would surely make it clearer to the users.
We could also easily define the behavior of zero values for a given parameter means the action "does not get applied at all" this way.

versions/1.1.0-dev.md
ralfhandl: I like that: parameters are global, and need to be referenced by actions that use them.

versions/1.1.0-dev.md
ralfhandl: Call,it "unintended abuse" or "oops". Writing down a linear list of multi-valued parameters that at are unfolded into nested loops with exponential runtime seems like an easy mistake to make.

Explicit loops in a wrapper script calling the overlay processor with single-valued environment variables does the same, is less error-prone, and reduces the implementation effort for overlay processors.

Do one thing, and do it well.

versions/1.1.0-dev.md
ralfhandl: I would prefer to make this feature even less complex:
- parameters are single-valued
- no implicit loops or "matrices"

Whoever needs loops and matrices can do them in a wrapper script around the overlay processor; practically all scripting languages have built-in loops.

ralfhandl:
I really want to restrict parameters to a single value and avoid the complexity of matrix processing.

That can be done in a wrapper script.

baywet:
I'll try to provide more context about the intention of this proposal. It was put together mainly for the matrix feature because it's the only use case we have for parameters.
I added support for environment variables as a nice to have to try to solve a similar need in a consistent manner. 
Yes, templating of actions can be done outside of the overlay, but it's added complexity and custom infrastructure that needs to be built for everyone who needs it. 
Removing the matrix aspect makes inline values definitions moot as well. 

I don't think I'll push this proposal to completion without matrix in one form or another. 

So if you instead have suggestions to mitigate the risks you outlined without removing the feature all together, I suggest we follow that path instead.

ralfhandl:
> Removing the matrix aspect makes inline values definitions moot as well.

That would be another nice simplification:
- parameter values always come from the environment 
- optionally allow a default value inline in case the environment variable is undefined

ralfhandl:
To me it seems extremely unlikely that the `target` expression and the `update` value will be parameterized with the same single string value, which would be necessary for the current implicit loops over multi-valued parameters. 

If you have three or more realistic examples where this is the case, I may reconsider.

Until then I believe that most use cases for loops over parameterized actions will require groups of parameter values.

Assume an API has paths for several "thingies" and we need to insert systematic descriptions for each operation on each thingy.

API structure is something like

~~~yaml
paths:
  /foo:
    get: ...
    patch: ...
  /bar:
    get: ...
    patch: ...
~~~

The descriptions we want to add are along the lines of
- Read Foobaroo thingies
- Update Foobaroo thingies
- Read Barberix thingies
- Update Barberix thingies

We would need an array of (thingy path segment, thingy name) 
- foo, Foobaroo
- bar, Barberix

and an array of (method, operation purpose)
- get, Read
- patch, Update

Then two nested loops over these arrays can produce the necessary matrix of actions.

Which means I need arrays of "structured" parameters.

baywet:
Thank you for the additional information.

If I'm following correctly, you'd essentially like to define the defaultValues as 

```typescript
string | Record<string, string>
```

correct?

I'm open to that, what does that mean for environment variables parsing though?

versions/1.1.0-dev.md
```diff
@@ -141,6 +142,25 @@ The properties of the resolved `copy` object MUST be compatible with the target
...
+### Parameter Object
+
+The object represents parameters that can be used with actions.
+
+#### Fixed Fields
+
+| Field Name | Type | Description |
+| ---- | :----: | ---- |
+| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique for the action the parameter is defined. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
+| <a name="parameter-source"></a>source | `inline` or `environment` | **REQUIRED** Defines where the values for the parameters should be retrieved from. When `inline` is set, values are retrieved from the `values` property. When `environment` is set, the values are retrieved by the tool from environment variables. |
```
SensibleWood: I'm wondering here whether their is merit in pulling environment variable up a level, into a parameter "type" with a set of semantics. This feels like a worthy exercise in terms of consistency across implementations in tooling i.e. provide guidance to tooling makers on env variable handling, etc. Not critical for this feature I'm sure, but might be useful as a building block for the future.

versions/1.1.0-dev.md
```diff
@@ -118,6 +118,7 @@ This object represents one or more changes to be applied to the target document
...
 | <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects object nodes, the value of this field MUST be an object with the properties and values to merge with each selected object. If the `target` selects array nodes, the value of this field MUST be an array to concatenate with each selected array, or an object or primitive value to append to each selected array. If the `target` selects primitive nodes, the value of this field MUST be a primitive value to replace each selected node. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that each of the target nodes MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-parameters"></a> | [[Parameter Object](#parameter-object)][] | An array of parameter objects to use and expand the action definition. MAY contain zero or more parameters. Parameters MUST have unique names across the collection. Default `undefined`. |
```
```suggestion
| <a name="action-parameters"></a> | [[Parameter Object](#parameter-object)] | An array of parameter objects to use and expand the action definition. MAY contain zero or more parameters. Parameters MUST have unique names across the collection. Default `undefined`. |
```

versions/1.1.0-dev.md
```diff
@@ -134,6 +135,31 @@ The properties of the `update` or `copy` object MUST be compatible with the targ
...
 
+### Parameter Object
+
+The object represents parameters that can be used with actions.
+
+#### Fixed Fields
+
+| Field Name | Type | Description |
+| ---- | :----: | ---- |
+| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique for the action the parameter is defined. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
```
```suggestion
| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique per action. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
```

versions/1.1.0-dev.md
```diff
@@ -134,6 +135,31 @@ The properties of the `update` or `copy` object MUST be compatible with the targ
...
+| <a name="parameter-default-values"></a>defaultValues | `string[] or Record<string, string>[]` | The values to use in case the environment variable is not defined. MAY contain zero or more entries. When values contains multiple entries, the action is effectively projected for each of the entries of the parameter. If multiple parameters are defined for any given action, parameters effectively act as a matrix definition and the action should be projected across each dimension. |
```
```suggestion
| <a name="parameter-default-values"></a>defaultValues | [`string`] or [Map(`string`, `string`] | The values to use in case the environment variable is not defined. MAY contain zero or more entries. When values contains multiple entries, the action is effectively projected for each of the entries of the parameter. If multiple parameters are defined for any given action, parameters effectively act as a matrix definition and the action should be projected across each dimension. |
```

versions/1.1.0-dev.md
```diff
@@ -134,6 +135,31 @@ The properties of the `update` or `copy` object MUST be compatible with the targ
...
+### Parameter Object
+
+The object represents parameters that can be used with actions.
+
+#### Fixed Fields
+
+| Field Name | Type | Description |
+| ---- | :----: | ---- |
+| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique per action. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
+| <a name="parameter-default-values"></a>defaultValues | [`string`] or [Map(`string`, `string`] | The values to use in case the environment variable is not defined. MAY contain zero or more entries. When values contains multiple entries, the action is effectively projected for each of the entries of the parameter. If multiple parameters are defined for any given action, parameters effectively act as a matrix definition and the action should be projected across each dimension. |
```
mikekistler, line 147: Seems like there is no way to say "don't look in the environment, just use these values". Is that intentional?

baywet: That was one reservation I had as well which is why I initially had the source enum in place.
Happy to restore it. Or maybe we can have a boolean instead to ignore the environment variable.

versions/1.1.0-dev.md
ralfhandl: Are we going to make parameters global and use/reference/import them in the actions?

If the values are provided in env variables, their names are global anyway, and same name means same value(s).

baywet: I'm open to it. What do we want the references to look like? we could use the $ref syntax so it feels familiar, and it opens the door to having more kinds of references, external references, etc... in the future if we need to. But the downside is that it's a bit verbose and might bring in much more than we need (again, external references, etc...)

versions/1.1.0-dev.md
baywet: @ralfhandl I'd like your input on this one please since you helped me refactor into a simpler syntax, we kind of lost a capability in the process.

versions/1.1.0-dev.md
ralfhandl: I wouldn’t use `$ref` and simply reference by name:
- global `parameters` is an array of parameter definition objects
- action-local `parameters` is an array of parameter names

Similar to the `tags` keyword in OpenAPI.

Or use different keywords, for example `usedParameters` in actions.

lornajane:
I am not in favour of the matrix feature. I would like us to consider simple parameters though.

I'm very confused about how we're using environment variables here, or it might be a terminology problem! Environment variables: the values in the environment when the tool executes? Or something else in this context?

My previous proposal for supporting environment variables was rejected because we received feedback about the difficulty of parsing dollar signs in value expressions where there might already be dollar signs in the OpenAPI content for other reasons. I notice however that we have again used dollar signs in the syntax and would like to make sure that we reach out to the wider community for their feedback before pushing through a change like this. /cc @ThomasRooney

baywet:
@lornajane why are you not in favour of matrix parameters?

As for the syntax, happy to change it to something else if we have a better suggestion

versions/1.1.0-dev.md
```diff
@@ -536,6 +562,61 @@ paths:
...
+
+This example shows how action parameters can be used to define multiple actions from a single one.
+
+```yaml
+overlay: 1.1.0
+info:
+  title: Makes a matrix projection
+  version: 1.0.0
+actions:
+  - description: Adds component schema ${foo}-${bar.description}
```
baywet, line 575:
```suggestion
  - description: Adds component schema #{foo}-#{bar.description}
```

baywet, line 575: comment from @mikeschinkel on the call: If we're worried about this string replace syntax, let's come up with edge cases where it might break. This way I could add those test cases to my experimental implementation here https://github.com/BinkyLabs/openapi-overlays-dotnet/pull/144

mikeschinkel: @baywet — Since there was a comment on the call about being unclear what _"interpolation"_ mean, maybe link to this in your markdown? https://en.wikipedia.org/wiki/String_interpolation

baywet:
@mikekistler @mikeschinkel @lornajane @ralfhandl I just rewrote the initial comment of this pull request to help clarify the vision, approach, remaining questions, etc... Please have a read, hopefully this helps us unlock further progress.

versions/1.1.0-dev.md
```diff
@@ -134,6 +135,31 @@ The properties of the `update` or `copy` object MUST be compatible with the targ
...
 
+### Parameter Object
+
+The object represents parameters that can be used with actions.
+
+#### Fixed Fields
+
+| Field Name | Type | Description |
+| ---- | :----: | ---- |
+| <a name="parameter-name"></a>name | `string` | **REQUIRED** A name for the parameter that is unique per action. The name MUST match the following regular expression `[a-zA-Z-0-9]+`. |
```
ralfhandl, line 146: Add note that same name in different actions means same environment variable value in all actions.

Which also makes different default values questionable.

versions/1.1.0-dev.md
baywet: thanks! added in the updated initial comment for this PR

versions/1.1.0-dev.md
```diff
@@ -134,6 +135,31 @@ The properties of the `update` or `copy` object MUST be compatible with the targ
 
 This object MAY be extended with [Specification Extensions](#specification-extensions).
 
+### Parameter Object
+
+The object represents parameters that can be used with actions.
```
mikekistler: "parameters" or "a parameter" ?

versions/1.1.0-dev.md
```diff
@@ -118,6 +118,7 @@ This object represents one or more changes to be applied to the target document
...
 | <a name="action-copy"></a>copy | `string` | A JSONPath expression selecting a single node to copy into the `target` nodes. If the `target` selects object nodes, the value of this field MUST be an object with the properties and values to merge with each selected object. If the `target` selects array nodes, the value of this field MUST be an array to concatenate with each selected array, or an object or primitive value to append to each selected array. If the `target` selects primitive nodes, the value of this field MUST be a primitive value to replace each selected node. This field has no impact if the `remove` field of this action object is `true` or if the `update` field contains a value. |
 | <a name="action-remove"></a>remove | `boolean` | A boolean value that indicates that each of the target nodes MUST be removed from the the map or array it is contained in. The default value is `false`. |
+| <a name="action-parameters"></a> | [[Parameter Object](#parameter-object)] | An array of parameter objects to use and expand the action definition. MAY contain zero or more parameters. Parameters MUST have unique names across the collection. Default `undefined`. |
```
mikekistler: Aren't you missing the field name here?
```suggestion
| <a name="action-parameters"></a>parameters | [[Parameter Object](#parameter-object)] | An array of parameter objects to use and expand the action definition. MAY contain zero or more parameters. Parameters MUST have unique names across the collection. Default `undefined`. |
```

versions/1.1.0-dev.md
mikekistler: Another possible approach, which I admit adds some complexity, but also a lot of flexibility, is to specify a value (or values) for the parameter with an "expression", similar to what is done for Link Objects in OpenAPI.

https://spec.openapis.org/oas/v3.1.1.html#runtime-expressions

versions/1.1.0-dev.md
```suggestion
The object represents a parameter that can be used with actions.
```

baywet:
@smikulcik I think the scope definition is a good topic, thank you for bringing this up. If you have a look at #33 you'll see it's an old issue that originated from the OpenAPI spec repository back in **2017** in fact. And it was moved to this one, as far as I understand because the TSC at the time didn't want to bring this complexity in the "root" specification.

I don't think we'll have much appetite from that group to bring this topic back for a potential 3.3 release. Additionally, the release cadence for OpenAPI is usually slower, because it's a more adopted and longer spec most likely, so we'd be putting this ability off at least one year the way I see it.

Here we have the opportunity to add something that makes sense with the rest of the feature set, that's been asked for in the ecosystem for a while, and get it out there in people's hands faster than in the root specification. This is one reason why I had started that work here.

As for a conditional expression syntax, even though this would be a logical follow up to introducing environment variables, it's a MUCH higher complexity than what is being proposed here. Even if we account for a super "simple" syntax (exact match, starts/ends with, contains, boolean expressions,...). Even if we adopted some normalized syntax, like the [RFC7644 filtering syntax](https://www.rfc-editor.org/rfc/rfc7644#section-3.4.2.2) or [CAML query](https://learn.microsoft.com/en-us/sharepoint/dev/schema/query-schema) 🤮, it'd still put a great burden on implementers.

And to answer the initial question: I personally see overlays as a transformation medium, a little like XSLT to XML, but more digestible. I know *you* use it as a diff, and it surely can be used this way as well, but I'm not in favour of limiting it to that aspect. As a matter of fact, the copy proposal that was merged in v1.1 would go against a "diff only" purpose.

nasha4:
I hope this thread is an appropriate place for this feedback.  I would like to opine as an overlay user that serving as simple diffs is not adequate.  One use case I have run into and been frustrated with the limitations of overlays, is working with a large OpenAPI spec that uses only "default" responses for all endpoints, instead of "200"s.  Ok, an overlay should be a slam dunk solution to this issue, right?  Well, it *can* be made to work:
```YAML
  - target: $.paths.*[?(@.responses.default && !@.responses['200'])].responses
    update: { "200": {} }
  - target: $.paths['/certificate-types'].get.responses['200']
    copy:   $.paths['/certificate-types'].get.responses.default
  - target: $.paths['/states'].get.responses['200']
    copy:   $.paths['/states'].get.responses.default
  - target: $.paths['/submission-types'].get.responses['200']
    copy:   $.paths['/submission-types'].get.responses.default
    # ... hundreds more lines like these
```
But I would really love to just be able to simply express something like:
```YAML
  - target: $.paths.*[?(@.responses.default && !@.responses['200'])].responses
    update: { "200": {} }
  - target: $.paths.*.*.responses['200']
    copy:   $.paths.*.*.responses.default
```
This is impossible with the limitations of the copy selector.

mikekistler:
@nasha4 This seems like a great use case if we could solve it with parameters.

@baywet Could this be done with the current proposal?  If so, what would that look like?

PR #265, opened 2025-12-16
docs: adds a clarification for different formats across documents

### LINKED ISSUE #260: Input/output Overlay / OAD document formats (JSON vs YAML)
In PR #246 (fixing issue #236) @baywet asked:

> Should we have a note to say that the format of the OpenAPI and Overlay documents MAY be different?

Now that I understand more of what he was asking, my answer is:

Conceptually, the format (meaning JSON vs JSON-compatible YAML) of each Overlay, each input OpenAPI Description document, and each output OpenAPI Description document are all completely independent.  You can mix and match JSON and JSON-compatible YAML arbitrarily.

My understanding is that tools apply one Overlay document to one OAD document at a time, so this ends up being pretty simple:  You need to allow those two input documents to be either format, and you can output the resulting OAD document in either format.

We should make all of that clear.  It's not entirely clear to me how much we need to say all of these options MUST be supported, since JSON <-> YAML coversions aren't hard, but... JSON <-> YAML conversions aren't hard, so integrating the right input/output support into tools isn't that burdensome either.

baywet:
fixes #260

### DIFF
--- a/versions/1.1.0-dev.md
+++ b/versions/1.1.0-dev.md
@@ -37,7 +37,7 @@ An Overlay document that conforms to the Overlay Specification is itself a JSON 
 
 Implementations MUST support applying either JSON or YAML Overlay documents to
 [[OpenAPI|OpenAPI]] descriptions
-represented in either YAML for JSON format.
+represented in either YAML or JSON format.
 
 All field names in the specification are **case sensitive**.
 This includes all fields that are used as keys in a map, except where explicitly noted that keys are **case insensitive**.


handrews (approved):

versions/1.1.0-dev.md
```diff
@@ -35,6 +35,8 @@ The Overlay Specification is versioned using a `major`.`minor`.`patch` versionin
 
 An Overlay document that conforms to the Overlay Specification is itself a JSON object, which may be represented either in [[RFC8259|JSON]] or [[YAML|YAML]] format. Examples in this specification will be shown in YAML for brevity.
 
+An Overlay document, whether represented in [[RFC8259|JSON]] or [[YAML|YAML]], MUST be applicable to an OpenAPI document represented in either [[RFC8259|JSON]] or [[YAML|YAML]]. Implementations MUST support application across differing formats.
```
DavidBiesack: The clause in first sentence is redundant with the previous sentence - I suggest removing it.
Stating the OpenAPI document is either JSON or YAML is not needed here; that is defined at https://spec.openapis.org/oas/v3.1.1#openapi-document.
I propose:
```
Implementations MUST support applying either JSON or YAML Overlay documents to
[OpenAPI documents](https://spec.openapis.org/oas/v3.1.1#openapi-document)
represented in either YAML for JSON format.
```

```suggestion
Implementations MUST support applying either JSON or YAML Overlay documents to
[[OpenAPI|OpenAPI]] descriptions
represented in either YAML for JSON format.
```

What do you think of this version?

DavidBiesack: Thanks, @baywet - this change looks good to me - clear, concise, and not overly restrictive.

DavidBiesack: Given that line 37 already has links `[[RFC8259|JSON]] or [[YAML|YAML]] format`, those same links may not be necessary in the revised lines on 38-39. Not major, but links can disrupt reading/comprehension. (There is research supporting _and_ refuting this assertion) -- but I think redundant links don't help and are not needed when it's the same thing being repeated in such close proximity. (That is why I did not propose repeating the links in my suggestion)

versions/1.1.0-dev.md
DavidBiesack, line 1: See @jeremyfiel's comment that this documentation should explicitly state that an Overlay document can be applied to any JSON/YAML document not just OAS documents

handrews:
@lornajane I agree, Overlays can become more powerful for the OpenAPI ecosystem by leveraging knowledge of OAS/Arazzo syntax and semantics.  JSON Patch and JSON Merge-Patch already exist as generic JSON modification systems.

@baywet there might be some subtleties here if "OpenAPI Description" (the "D" should be capitalized) is used, as opposed to just any OpenAPI document (little "d").  Applying to a doccument is intuitive.  But if we're applying to OADs, does that mean applying to an entry document applies to all documents reachable from that entry document?  (AFAICT Overlays currently work on individual documents, my apologies if I am misunderstanding).

DavidBiesack (approved):

versions/1.1.0-dev.md
```diff
@@ -35,6 +35,10 @@ The Overlay Specification is versioned using a `major`.`minor`.`patch` versionin
 
 An Overlay document that conforms to the Overlay Specification is itself a JSON object, which may be represented either in [[RFC8259|JSON]] or [[YAML|YAML]] format. Examples in this specification will be shown in YAML for brevity.
 
+Implementations MUST support applying either JSON or YAML Overlay documents to
+[[OpenAPI|OpenAPI]] descriptions
+represented in either YAML for JSON format.
```
baywet:
```suggestion
represented in either YAML or JSON format.
```

PR #266, opened 2025-12-16
DO NOT MERGE - release notes for v1.1

baywet:
Signed-off-by: Vincent Biret <vincentbiret@hotmail.com>

PR #276, opened 2026-01-14
v1.1.0 release

baywet:
this releases the 1.1.0 version of the Overlay specification following the contributing guidance.
depends on #275
closes #266


versions/1.1.0.md
earth2marsh, line 586: Very minor thought... "Documents," seems marginally better to me here than "Descriptions," since the comment may exist at the document level without affecting the description that the OpenAPI Document describes. (This seems pedantic enough that this observation could be ignored for this release even if folks were to agree.)

baywet: Thanks! If you're referring to the format addition, we've had extensive discussions about which term to use here in #265 for context

ralfhandl:
A nice step forward!

Special thanks to @baywet for adding `copy`, supporting the `update` round-off, and adding a .NET implementation!

PR #277, opened 2026-01-14
docs: adds WSL related precisions for scripts

baywet:
while working on #276 and using WSL, I ran into a few issues. This PR adds precisions for people running a similar setup

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -133,7 +133,7 @@ A new minor version X.(Y+1).0 or major version (X+1).0.0 is started similarly:
 
 If you are running those scripts using Windows Subsystems for Linux (WSL), and cloned the repository under windows, you'll need to make a few adjustments before you can run these procedures:
 
-1. Save the scripts using LF, not CRLF to avoid parsing issues. You can use VSCode or any other editor to do that.
+1. Save the scripts using LF, not CRLF to avoid parsing issues. You can use VSCode or any other editor to do that. Alternatively, you may clone the repository again from WSL to workaround the line return issue.
 1. Make sure you run the npm install from WSL and not from windows.
 1. If you run into issues launching chrome, [review this StackOverflow answer](https://stackoverflow.com/a/78776116/3808675).
 


CONTRIBUTING.md
```diff
@@ -129,6 +129,14 @@ A new minor version X.(Y+1).0 or major version (X+1).0.0 is started similarly:
    - commit changes
 2. Push branch `start-X'.Y'.0` and merge into `main` via pull request
 
+### Publishing using WSL
+
+If you are running those scripts using Windows Subsystems for Linux (WSL), and cloned the repository under windows, you'll need to make a few adjustments before you can run these procedures:
+
+1. Save the scripts using LF, not CRLF to avoid parsing issues. You can use VSCode or any other editor to do that.
```
ralfhandl: Git setting `core.autocrlf=input` should do that and is a good choice for Windows users contributing to a *nix project.

CONTRIBUTING.md
```diff
@@ -129,6 +129,14 @@ A new minor version X.(Y+1).0 or major version (X+1).0.0 is started similarly:
...
+1. Make sure you run the npm install from WSL and not from windows.
```
ralfhandl, line 137: Which is a reason to clone from WSL and avoid accessing a local repo from both operating systems. VS Code can be started with the `code` command from WSL.

CONTRIBUTING.md
baywet: I don't think asking people to change their local configuration to accommodate a script is going to be super welcoming. But I like your suggestion of making another clone.

baywet:
```suggestion
1. Save the scripts using LF, not CRLF to avoid parsing issues. You can use VSCode or any other editor to do that. Alternatively, you may clone the repository again from WSL to workaround the line return issue.
```

CONTRIBUTING.md
baywet: I think that with the local clone I've suggested in the previous step, it should help reduce the incidence of "cross platform installs" here. I think we should leave this suggestion just in case people make the same mistake though. Thoughts?

PR #322, opened 2026-04-13
feat: adds support for self in overlay

baywet:
fixes #310

heavy use of GCC to put this one together. (based on the discussion in the issue, and the arazzo PR as a template)

### DIFF
--- a/versions/1.2.0-dev.md
+++ b/versions/1.2.0-dev.md
@@ -67,32 +67,16 @@ Overlay Description authors SHOULD consider how text using such extensions will 
 
 ### Parsing Documents
 
-Each possible target [[OpenAPI]] Description document MUST be fully parsed in order to locate possible `extends` resolution targets before attempting to resolve an Overlay's `extends` value.
-Reference targets include the root object's `$self` field when present.
+Implementations MUST NOT treat an `extends` value as unresolvable before checking all possible target [[OpenAPI]] Description documents provided to the implementation.
 
-Implementations MUST NOT treat an `extends` value as unresolvable before completely parsing all possible target OpenAPI Description documents provided to the implementation.
+To ensure interoperability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value MUST match the target's `$self` URI.
+Implementations MUST examine all available target documents for matching `$self` values before concluding that no document matches the `extends` URI.
 
-#### Fragmentary Parsing
-
-**Fragmentary parsing** occurs when an implementation parses only the specific part of a target description that seems needed, rather than parsing complete candidate target descriptions first.
-This practice is **strongly discouraged** and produces undefined behavior.
-
-If only parts of a target description are parsed, implementations can miss the root `$self` field and incorrectly fail to resolve, or mis-resolve, `extends` values.
-When fragmentary parsing causes resolution failures or unintended matches, the resulting behavior is _undefined_ and _implementation-defined_.
-
-Implementations MUST parse complete candidate target descriptions before resolving `extends` values to ensure predictable and interoperable behavior.
-
-#### Identifier-Based Referencing
-
-To ensure interoperability, when an Overlay identifies a target [[OpenAPI]] Description by URI using `extends`, that URI MUST be the target description's `$self` URI if the target defines `$self`.
-
-This means implementations MUST examine possible target OpenAPI Description documents to check for matching `$self` values before deciding no document matches the `extends` URI.
-
-Implementations MAY support additional, non-interoperable lookup methods (such as retrieval URI matching), but relying on those methods is NOT RECOMMENDED.
+Implementations MAY also support retrieval URI matching, which can be useful when `$self` is not defined in the target document.
 
 ### Relative References in URIs
 
-URIs used in an Overlay document, including `$self` and `extends`, are resolved as _identifiers_.
+URIs used in an Overlay document, including `$self` and `extends`, are identifiers, and not necessarily locators (URLs).
 
 Unless specified otherwise, all fields that are URI references MAY be relative references as defined by [RFC3986 Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2).
 
@@ -108,7 +92,7 @@ The base URI for resolving relative references within an Overlay document is det
 
 #### Resolving URI Fragments
 
-If a URI contains a fragment identifier, then the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
+Where a URI appears in contexts such as CommonMark hyperlinks within `description` fields and contains a fragment identifier, the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
 
 For Overlay Object fields that identify whole documents (`$self` and `extends`), fragment identifiers MUST NOT be used.
 
@@ -129,7 +113,7 @@ This is the root object of the [Overlay](#overlay).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="overlay-version"></a>overlay | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the Overlay Specification that the Overlay document uses. The `overlay` field SHOULD be used by tooling to interpret the Overlay document. |
-| <a name="overlay-self"></a>$self | `string` | A URI-reference for the Overlay document. This string MUST be in the form of a URI-reference as defined by [RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). When present, this field provides the self-assigned URI of this Overlay document, which also serves as its base URI in accordance with [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1) for resolving relative references within this document. The `$self` URI MUST NOT contain a fragment identifier. Overlay documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution. |
+| <a name="overlay-self"></a>$self | `string` | A URI-reference for the Overlay document. This string MUST be in the form of a URI-reference as defined by [RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). When present, this field provides the self-assigned URI of this Overlay document, which also serves as its base URI in accordance with [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1) for resolving relative references within this document. The `$self` URI MUST NOT contain a fragment identifier. Overlay documents MAY include a `$self` field to ensure portable, unambiguous reference resolution. |
 | <a name="overlay-info"></a>info | [Info Object](#info-object) | **REQUIRED**. Provides metadata about the Overlay. The metadata MAY be used by tooling as required. |
 | <a name="overlay-extends"></a>extends | `string` | URI reference that identifies the target document (such as an [[OpenAPI]] document) this overlay applies to. |
 | <a name="overlay-actions"></a>actions | [[Action Object](#action-object)] | **REQUIRED** An ordered list of actions to be applied to the target document. The array MUST contain at least one value. |
@@ -141,10 +125,18 @@ The list of actions MUST be applied in sequential order to ensure a consistent o
 The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] description.
 Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI description(s).
 
-The `extends` property is an identifier.
-When candidate OpenAPI description documents define `$self`, the `extends` value MUST be matched to `$self` as described in [Identifier-Based Referencing](#identifier-based-referencing).
+In the following example, the Overlay identifies a target OpenAPI description using an absolute URI:
 
-In the following example the Overlay document includes `$self` and identifies a target OpenAPI description with an absolute URI:
+```yaml
+overlay: 1.2.0
+info:
+  title: Overlay for the Tic Tac Toe API document
+  version: 1.0.0
+extends: 'https://api.example.com/openapi/tictactoe.yaml'
+...
+```
+
+In the following example, the Overlay document also includes `$self`, which establishes the document's identity and base URI:
 
 ```yaml
 overlay: 1.2.0
@@ -644,7 +636,15 @@ Interoperable Overlay Documents MUST use RFC9535 JSONPath query expressions and 
 
 Some formats, like [YAML](https://yaml.org/) or [JSONC](https://jsonc.org/), support using comments which do not impact the semantic meaning of the description. Applying Overlay Actions to a description MAY result in the loss of such comments in the updated description. The exact behavior is specific to the tool implementing the Overlay Specification.
 
-## Appendix A: Examples of Base URI Determination and Identifier Resolution
+## Appendix A: Revision History
+
+| Version | Date | Notes |
+| ---- | ---- | ---- |
+| 1.2.0 | TBD | Release of the Overlay Specification 1.2.0 |
+| 1.1.0 | 2026-01-14 | Release of the Overlay Specification 1.1.0 |
+| 1.0.0 | 2024-10-17 | First release of the Overlay Specification |
+
+## Appendix B: Examples of Base URI Determination and Identifier Resolution
 
 ### Base URI Within Content (Using `$self`)
 
@@ -683,7 +683,7 @@ actions:
 
 Retrieved from `file:///Users/dev/projects/overlays/purchase.overlay.yaml`, the relative `extends` URI resolves to `file:///Users/dev/projects/openapi/petstore.yaml`.
 
-### Identity vs Location: Why `extends` Should Match `$self`
+### Why Potential Target Documents Should Be Checked for `$self`
 
 An Overlay may be given this `extends` value:
 
@@ -711,11 +711,3 @@ paths: {}
 ```
 
 Because `extends` is identifier-based, the match succeeds using the target description's `$self` URI, even though retrieval occurred from a different location.
-
-## Appendix B: Revision History
-
-| Version | Date | Notes |
-| ---- | ---- | ---- |
-| 1.2.0 | TBD | Release of the Overlay Specification 1.2.0 |
-| 1.1.0 | 2026-01-14 | Release of the Overlay Specification 1.1.0 |
-| 1.0.0 | 2024-10-17 | First release of the Overlay Specification |
--- a/versions/1.2.0-dev.md
+++ b/versions/1.2.0-dev.md
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown 
 While the framing of CommonMark 0.27 as a minimum requirement means that tooling MAY choose to implement extensions on top of it, note that any such extensions are by definition implementation-defined and will not be interoperable.
 Overlay Description authors SHOULD consider how text using such extensions will be rendered by tools that offer only the minimum support.
 
+### Parsing Documents
+
+Each possible target [[OpenAPI]] Description document MUST be fully parsed in order to locate possible `extends` resolution targets before attempting to resolve an Overlay's `extends` value.
+Reference targets include the root object's `$self` field when present.
+
+Implementations MUST NOT treat an `extends` value as unresolvable before completely parsing all possible target OpenAPI Description documents provided to the implementation.
+
+#### Fragmentary Parsing
+
+**Fragmentary parsing** occurs when an implementation parses only the specific part of a target description that seems needed, rather than parsing complete candidate target descriptions first.
+This practice is **strongly discouraged** and produces undefined behavior.
+
+If only parts of a target description are parsed, implementations can miss the root `$self` field and incorrectly fail to resolve, or mis-resolve, `extends` values.
+When fragmentary parsing causes resolution failures or unintended matches, the resulting behavior is _undefined_ and _implementation-defined_.
+
+Implementations MUST parse complete candidate target descriptions before resolving `extends` values to ensure predictable and interoperable behavior.
+
+#### Identifier-Based Referencing
+
+To ensure interoperability, when an Overlay identifies a target [[OpenAPI]] Description by URI using `extends`, that URI MUST be the target description's `$self` URI if the target defines `$self`.
+
+This means implementations MUST examine possible target OpenAPI Description documents to check for matching `$self` values before deciding no document matches the `extends` URI.
+
+Implementations MAY support additional, non-interoperable lookup methods (such as retrieval URI matching), but relying on those methods is NOT RECOMMENDED.
+
 ### Relative References in URIs
 
-Unless specified otherwise, all fields that are URI references MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+URIs used in an Overlay document, including `$self` and `extends`, are resolved as _identifiers_.
+
+Unless specified otherwise, all fields that are URI references MAY be relative references as defined by [RFC3986 Section 4.2](https://tools.ietf.org/html/rfc3986#section-4.2).
+
+#### Establishing the Base URI
+
+Relative URI references are resolved using the appropriate base URI, which MUST be determined in accordance with [RFC3986 Section 5.1.1 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.1).
+
+The base URI for resolving relative references within an Overlay document is determined as follows:
+
+- If the [`$self`](#overlay-self) field is present and is an absolute URI, the base URI is the `$self` URI (per [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1): Base URI Embedded in Content).
+- If the [`$self`](#overlay-self) field is present and is a relative URI-reference, it MUST first be resolved against the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2) before being used as the base URI for resolving other relative references.
+- If the [`$self`](#overlay-self) field is not present, the base URI MUST be determined from the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2). The most common base URI source in this case is the retrieval URI of the Overlay document (per [RFC3986 Section 5.1.3](https://tools.ietf.org/html/rfc3986#section-5.1.3)), though other sources such as encapsulating entities ([RFC3986 Section 5.1.2](https://tools.ietf.org/html/rfc3986#section-5.1.2)) or application-specific defaults ([RFC3986 Section 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.4)) MAY apply.
+
+#### Resolving URI Fragments
+
+If a URI contains a fragment identifier, then the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
+
+For Overlay Object fields that identify whole documents (`$self` and `extends`), fragment identifiers MUST NOT be used.
+
+#### Relative URI References in CommonMark Fields
+
+Relative references in CommonMark hyperlinks (such as those in `description` fields) are resolved in their rendered context, which might differ from the context of the Overlay document.
 
 ### Schema
 
@@ -82,6 +129,7 @@ This is the root object of the [Overlay](#overlay).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="overlay-version"></a>overlay | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the Overlay Specification that the Overlay document uses. The `overlay` field SHOULD be used by tooling to interpret the Overlay document. |
+| <a name="overlay-self"></a>$self | `string` | A URI-reference for the Overlay document. This string MUST be in the form of a URI-reference as defined by [RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). When present, this field provides the self-assigned URI of this Overlay document, which also serves as its base URI in accordance with [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1) for resolving relative references within this document. The `$self` URI MUST NOT contain a fragment identifier. Overlay documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution. |
 | <a name="overlay-info"></a>info | [Info Object](#info-object) | **REQUIRED**. Provides metadata about the Overlay. The metadata MAY be used by tooling as required. |
 | <a name="overlay-extends"></a>extends | `string` | URI reference that identifies the target document (such as an [[OpenAPI]] document) this overlay applies to. |
 | <a name="overlay-actions"></a>actions | [[Action Object](#action-object)] | **REQUIRED** An ordered list of actions to be applied to the target document. The array MUST contain at least one value. |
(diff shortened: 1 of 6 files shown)


versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+- If the [`$self`](#overlay-self) field is not present, the base URI MUST be determined from the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2). The most common base URI source in this case is the retrieval URI of the Overlay document (per [RFC3986 Section 5.1.3](https://tools.ietf.org/html/rfc3986#section-5.1.3)), though other sources such as encapsulating entities ([RFC3986 Section 5.1.2](https://tools.ietf.org/html/rfc3986#section-5.1.2)) or application-specific defaults ([RFC3986 Section 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.4)) MAY apply.
+
+#### Resolving URI Fragments
+
+If a URI contains a fragment identifier, then the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
+
+For Overlay Object fields that identify whole documents (`$self` and `extends`), fragment identifiers MUST NOT be used.
```
baywet: Here I find the wording a bit confusing. The two statements *seem* to contradict each other. I think we should qualify the first one further to state things like common mark links etc... Thoughts?

versions/1.2.0-dev.md
```diff
@@ -90,29 +138,38 @@ This object MAY be extended with [Specification Extensions](#specification-exten
...
+In the following example the Overlay document includes `$self` and identifies a target OpenAPI description with an absolute URI:
 
 ```yaml
 overlay: 1.2.0
+$self: 'https://example.com/overlays/tictactoe.overlay.yaml'
 info:
   title: Overlay for the Tic Tac Toe API document
   version: 1.0.0
-extends: 'https://raw.githubusercontent.com/OAI/learn.openapis.org/refs/heads/main/examples/v3.1/tictactoe.yaml'
+extends: 'https://api.example.com/openapi/tictactoe.yaml'
```
baywet, line 144: Here because of the changes, we effectively lost the "regular resolution" example (use of extend without self). I suggest we re-introduce it. Even though we have some in the appendix section, it's a long way down. Thoughts?

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
 While the framing of CommonMark 0.27 as a minimum requirement means that tooling MAY choose to implement extensions on top of it, note that any such extensions are by definition implementation-defined and will not be interoperable.
 Overlay Description authors SHOULD consider how text using such extensions will be rendered by tools that offer only the minimum support.
 
+### Parsing Documents
+
+Each possible target [[OpenAPI]] Description document MUST be fully parsed in order to locate possible `extends` resolution targets before attempting to resolve an Overlay's `extends` value.
```
handrews: The requirement here is _slightly_ different: You can't declare a referencing (including `extends`) un-resolvable without parsing all known potential target documents ***[NOTE: I later saw that you have this requirement correct further down, I'm leaving up this explanation to explain why this part shouldn't be here, apologies if it is overkill].***

To explain this, let's assume we (for... reasons) have some OADs and overlays that have been placed on an air-gapped network.  The laws of physics prevent fetching them from any network location- no wires, no wireless, no nothing.

The Overlay tool would need to be told what documents to consider.  It might be given the intended retrieval URIs of the documents as well, meaning "pretend that you retrieved these documents from these URIs".  This is common in JSON Schema reference resolvers, because you are not guaranteed to have an `$id`, and you are not guaranteed to have full network or filesystem access.

If one of those retrieval URIs matches your `$extends`, you can resolve it to that retrieval URI.  Hopefully the `$self` matches, but it technically does not have to.  A mismatch would be... not a user error, but a user being weird and counter-intuitive.

On the other hand, if you weren't provided a retrieval URI for each document (or if none of them match your `extends`) _and_ the `extends` field is a network URI (on your air-gapped machine with no network access, because physics), then your implementation needs to have been told what documents to check, and needs to search them for a matching `$self`.

So the condition here is "don't give up until you've parsed everything".  We definitely _do not_ want to make parsing every possible target a _precondition_ to resolution.  That might be a lot of wasted effort!

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+Reference targets include the root object's `$self` field when present.
```
handrews: There's something a bit odd about this phrasing to me.  `$self` provides a self-assigned URI.  URIs are what are used for references. Saying "reference targets include the root object's `$self`" makes this feel like a special case.  It's not.  It's just a consequence of two steps:

1.  References use URIs
2. `$self` assignes a URI

The reference (`extends`) shouldn't care at all where the URI came from.  In fact, it should be possible to write a URI determination module and a reference resolution module independently.  The URIs are the interface, how they got there is an implementation detail that only applies to one side.  Does that make sense?

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
+Implementations MUST NOT treat an `extends` value as unresolvable before completely parsing all possible target OpenAPI Description documents provided to the implementation.
+
+#### Fragmentary Parsing
+
+**Fragmentary parsing** occurs when an implementation parses only the specific part of a target description that seems needed, rather than parsing complete candidate target descriptions first.
+This practice is **strongly discouraged** and produces undefined behavior.
```
handrews: Has this behavior previously been recommended in Overlays?  If not, I would cut this whole section.  The only reason anything like this is in the OAS is that OAS 2.0 specified fragmentary parsing.  And OAS 3.0 was completely unclear.  So we needed to explain the change.  But if Overlays only point to entire documents, there's no need for this and it's just confusing.

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
+Implementations MUST parse complete candidate target descriptions before resolving `extends` values to ensure predictable and interoperable behavior.
+
+#### Identifier-Based Referencing
+
+To ensure interoperability, when an Overlay identifies a target [[OpenAPI]] Description by URI using `extends`, that URI MUST be the target description's `$self` URI if the target defines `$self`.
+
+This means implementations MUST examine possible target OpenAPI Description documents to check for matching `$self` values before deciding no document matches the `extends` URI.
+
+Implementations MAY support additional, non-interoperable lookup methods (such as retrieval URI matching), but relying on those methods is NOT RECOMMENDED.
```
handrews: I don't think this needs to be separate from the first section, which overlaps considerably.

I've probably said contradictory things about using retrieval URIs.  I don't want to be too discouraging here, because they do work, and sometimes they are the _only_ option.  If a tool handles them, it should be interoperable.  But we don't say exactly how a tool learns the retrieval URI (provided by user, determined by actually fetching it as a URL, etc.).  And the caveats also apply to URIs that might come from other contexts, e.g. `Content-Location` in `multipart/related`, although I'm not aware of anyone actually doing that (which is a shame, it's a nice solution to bundling documents of different formats).

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,56 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
 ### Relative References in URIs
 
-Unless specified otherwise, all fields that are URI references MAY be relative references as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-4.2).
+URIs used in an Overlay document, including `$self` and `extends`, are resolved as _identifiers_.
```
handrews: I'm not sure most people know what "resolved as _identifiers_" means.  I tend to say something like "are identifiers (URIs), and not necessarily locators (URLs)."  This gets paired with discussion of how documents can be made available to the implementation so that the identifiers are seen.

versions/1.2.0-dev.md
handrews: Do Overlays ever use fragment references?  Perhaps in Schema Objects?  Or is it always only whole-document external references or same-document internal references?

I also still think that talking about forbidding fragments for `extends` doesn't make much sense.  `$self` already forbids it.  A document retrieval URI won't have one, unless _maybe_ if the Overlay were embedded in a larger JSON-like document, but that's outside of the spec and we can ignore it.

Under what circumstances would someone try to use a fragment in `extends`?  What would they expect to happen?  I'm trying to understand why we need to say anything here.

versions/1.2.0-dev.md
```diff
@@ -82,6 +129,7 @@ This is the root object of the [Overlay](#overlay).
 | Field Name | Type | Description |
 | ---- | :----: | ---- |
 | <a name="overlay-version"></a>overlay | `string` | **REQUIRED**. This string MUST be the [version number](#versions) of the Overlay Specification that the Overlay document uses. The `overlay` field SHOULD be used by tooling to interpret the Overlay document. |
+| <a name="overlay-self"></a>$self | `string` | A URI-reference for the Overlay document. This string MUST be in the form of a URI-reference as defined by [RFC3986 Section 4.1](https://tools.ietf.org/html/rfc3986#section-4.1). When present, this field provides the self-assigned URI of this Overlay document, which also serves as its base URI in accordance with [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1) for resolving relative references within this document. The `$self` URI MUST NOT contain a fragment identifier. Overlay documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution. |
```
handrews: > Overlay documents SHOULD include a `$self` field to ensure portable, unambiguous reference resolution.

I had not noticed this in the Arazzo PR.  I just put the following comment there and it applies here, too, perhaps even more so if `extends` is the only field that would use `$self` right now, and it is an optional field:

> While [including `$self`] is my personal preference, there are times when you really do want things to be relative to wherever you are, so I hesitate to include a SHOULD here. A SHOULD is pretty strong.

versions/1.2.0-dev.md
```diff
@@ -90,29 +138,38 @@ This object MAY be extended with [Specification Extensions](#specification-exten
 
 The list of actions MUST be applied in sequential order to ensure a consistent outcome. Actions are applied to the result of the previous action. This enables objects to be deleted in one action and then re-created in a subsequent action, for example.
 
-The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] document. Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI document(s).
+The `extends` property can be used to indicate that the Overlay was designed to update a specific [[OpenAPI]] description.
+Where no `extends` is provided it is the responsibility of tooling to apply the Overlay document to the appropriate OpenAPI description(s).
+
+The `extends` property is an identifier.
+When candidate OpenAPI description documents define `$self`, the `extends` value MUST be matched to `$self` as described in [Identifier-Based Referencing](#identifier-based-referencing).
```
handrews: This feels a bit redundant, and as noted before I'm not entirely sure it ought to be called out anyway as it's just a consequence of how URIs work.

versions/1.2.0-dev.md
handrews: Well... when `extends` is a full URI (starting with a scheme) the base URI is irrelevant, so this `extends` example behaves the same regardless of `$self`.  So I'm not sure what was lost?

versions/1.2.0-dev.md
```diff
@@ -587,7 +644,75 @@ Interoperable Overlay Documents MUST use RFC9535 JSONPath query expressions and
 
 Some formats, like [YAML](https://yaml.org/) or [JSONC](https://jsonc.org/), support using comments which do not impact the semantic meaning of the description. Applying Overlay Actions to a description MAY result in the loss of such comments in the updated description. The exact behavior is specific to the tool implementing the Overlay Specification.
 
-## Appendix A: Revision History
+## Appendix A: Examples of Base URI Determination and Identifier Resolution
```
handrews: I believe Revision History is always Appendex A in the OAS.  The others appendices come afterwards.

versions/1.2.0-dev.md
```diff
@@ -587,7 +644,75 @@ Interoperable Overlay Documents MUST use RFC9535 JSONPath query expressions and
...
+extends: ../openapi/petstore.yaml
+actions:
+  - target: '$.info'
+    update:
+      x-overlay: purchase
+```
+
+Retrieved from `file:///Users/dev/projects/overlays/purchase.overlay.yaml`, the relative `extends` URI resolves to `file:///Users/dev/projects/openapi/petstore.yaml`.
+
+### Identity vs Location: Why `extends` Should Match `$self`
```
handrews: I'd say this is more "why potential target documents should be checked for `$self`", although I don't feel strongly about this.

handrews (changes requested):

versions/1.2.0-dev.md
baywet: My concern is that we don't have an example *without* $self anymore and that might confuse people?

versions/1.2.0-dev.md
baywet: in v1.1 and 1.0: only whole document references (we want an entire OpenAPI document). I think some people target fragment documents (e.g. a single JSON schema), but recent discussions on the topic have lead to the conclusion that they really shouldn't be doing that.

in 1.2, we're introducing same document references for the reusable actions. Since there's already been a discussion about being able to refer to reusable actions that live outside of the overlay document, this might introduce external references with fragment identifiers in the future. We explicitly decided NOT to do that for now.

versions/1.2.0-dev.md
baywet: I've re-added an example without self to address this concern in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: merged the sections together in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: updated to MAY in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: ah... I thought I was improving readability of the document moving it down. Sorry. Reverted in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: updated to your suggestion in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: Removed in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: removed in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: updated with your suggestion in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
baywet: hopefully the rewording in afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d makes it much easier to understand without omitting important information. Let me know what you think.

versions/1.2.0-dev.md
baywet: I *believe* the rewrite of the section makes that a bit clearer. But I'm happy to iterate here now that we have less text. afd5fc33bdb7bbb13d5f4f29a3ec57b92251296d

versions/1.2.0-dev.md
handrews: > I think some people target fragment documents (e.g. a single JSON schema), but recent discussions on the topic have lead to the conclusion that they really shouldn't be doing that.

I think the place to handle this is under "Parsing Documents", with language requiring each document to be fully parsed.  That makes it clear that extracting only a fragment goes against the parsing model.  And yes, extracting a JSON Schema from an OAD without checking the `openapi`, `$self` and `jsonSchemaDialect` fields, as wall as any parent schema `$id` or `$schema` fields, leads to unpredictable behavior and possibly security vulnerabilities.

I think just keeping the first paragraph here and not the second should be fine.  The prohibitions on fragments in `$self` and `extends` flow from how they work already.  Even if yo uhave reusable actions elsewhere, that wouldnt' be via `extends`.

versions/1.2.0-dev.md
handrews: @baywet  it's interested that you softened that language with a "MAY" while @frankkilcommins made it a "can" over in Overlays.  I have no idea which is better and am fine with either, I just thought I'd make a note of it.

versions/1.2.0-dev.md
```diff
@@ -594,3 +643,71 @@ Some formats, like [YAML](https://yaml.org/) or [JSONC](https://jsonc.org/), sup
 | 1.2.0 | TBD | Release of the Overlay Specification 1.2.0 |
 | 1.1.0 | 2026-01-14 | Release of the Overlay Specification 1.1.0 |
 | 1.0.0 | 2024-10-17 | First release of the Overlay Specification |
+
+## Appendix B: Examples of Base URI Determination and Identifier Resolution
```
handrews, line 644: @baywet Is there a reason you skipped the enclosing context and application-assigned cases that we have in OAS and Arazzo?

versions/1.2.0-dev.md
baywet: To give you more context about my decision process: I feel like the more prescriptive we can be, the better as it removes ambiguity. MAY is normative to some level, at least more than can. It *felt like* providing additional guidance to me.

versions/1.2.0-dev.md
baywet: an oversight, thanks for calling this out, updated.

versions/1.2.0-dev.md
baywet: I forgot to reply to this comment, but the paragraph were merged and I believe this is all the changes we needed? Let me know if I missed anything.

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,40 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
 While the framing of CommonMark 0.27 as a minimum requirement means that tooling MAY choose to implement extensions on top of it, note that any such extensions are by definition implementation-defined and will not be interoperable.
 Overlay Description authors SHOULD consider how text using such extensions will be rendered by tools that offer only the minimum support.
 
+### Parsing Documents
+
+Implementations MUST NOT treat an `extends` value as unresolvable before checking all possible target [[OpenAPI]] Description documents provided to the implementation.
+
+To ensure interoperability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value MUST match the target's `$self` URI.
```
handrews: I apologize for not noticing this earlier, but I'm not sure "interoperability" is the ideal word here.

This is really more about portability (of the documents) than interoperability.  Given the same input, implementations would be expected to resolve the references the same way.  I suppose this is a bit more interoperable as implementations might have different ways to figure out the current document's location for resolving a relative `extends`, but... hmm... I think this is more about portability of document sets.  idk, does anyone else think "interoperability" is better here?

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,40 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+Implementations MUST examine all available target documents for matching `$self` values before concluding that no document matches the `extends` URI.
+
+Implementations MAY also support retrieval URI matching, which can be useful when `$self` is not defined in the target document.
```
handrews: I'm pretty sure that if there's no `$self` then whatever the past requirements for `extends` were would still apply.  Presumably that was a "MUST match based on the retrieval URI" or some similar location thing?

I'm guessing that this wording is based on reading this from the OAS:

> Implementations MAY support document retrieval, although see the [Security Considerations](https://spec.openapis.org/oas/v3.2.0.html#security-considerations) sections for additional guidance.

The MAY there is referring to _actually performing an I/O operation to fetch the document._  That is optional, and potentially a security issue so it really shouldn't be on by default.  But the rest of that paragraph basically says "or you handle this by letting the user tell you what the retrieval URI was."

versions/1.2.0-dev.md
```diff
@@ -594,3 +643,146 @@ Some formats, like [YAML](https://yaml.org/) or [JSONC](https://jsonc.org/), sup
...
+extends: specs/petstore.yaml  # Resolves using application default
+actions:
+  - target: '$.info'
+    update:
+      x-overlay: purchase
+```
+
+If the application assigns base URI `https://overlays.example.com/a7b3c4d5` to this document, then `specs/petstore.yaml` resolves to `https://overlays.example.com/specs/petstore.yaml`.
+
+**Note:** While a base URI of `https://overlays.example.com/` (with trailing slash) is technically valid per RFC3986, the final path component is an empty string. In practice, implementations typically assign unique identifiers as the final component to distinguish documents.
```
handrews: Reading this now, I kinda feel like it's just adding potential confusion.  If people don't think to try this, they don't need to think about it at all.  And if they do try it and have problems... well, they can look up how relative URI reference resolution works.

handrews (changes requested):
This is getting quite close!  BTW, I have only had time to review the markdown file, I hope someone else can look at the tests and schemas, and just ping me if there are questions you can't resolve.

versions/1.2.0-dev.md
baywet:
```suggestion
To ensure portability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value SHOULD match the target's `$self` URI.
```

versions/1.2.0-dev.md
baywet: so you'd essentially remove the note all together?

handrews: I would, yes. (the "\*\*Note:\*\*" paragraph)

versions/1.2.0-dev.md
handrews: I also think "MUST" is a bit weird with a conditional on it.  MUST is an absolute requirement, but we also allow this to work via location, even if that is not portable.  So I think this ought (OUGHT? 😜 ) to be a SHOULD.  I think?

versions/1.2.0-dev.md
baywet:
```suggestion
```

versions/1.2.0-dev.md
```diff
@@ -594,3 +643,146 @@ Some formats, like [YAML](https://yaml.org/) or [JSONC](https://jsonc.org/), sup
...
+  version: 1.0.0
+extends: specs/petstore.yaml  # Resolves using application default
+actions:
+  - target: '$.info'
+    update:
+      x-overlay: purchase
+```
+
+If the application assigns base URI `https://overlays.example.com/a7b3c4d5` to this document, then `specs/petstore.yaml` resolves to `https://overlays.example.com/specs/petstore.yaml`.
+
```
baywet, line 733:
```suggestion
```

versions/1.2.0-dev.md
baywet: I re-read your comment, and the paragraph, multiple times to ensure I was understanding things properly here.

I feel this paragraph is actually redundant, the extends behaviour is documented in details for URI retrieval. My suggestion is to remove this entirely. 

Essentially here is what we want people to do (in order, read like a truth table):
1. extends, no self, no contextual information => do uri retrievial with extends
2. extends, no self, contextual information => do uri retrieval with extends, then try the contextual information (tool specific)
3. no extends, no self, contextual information => do uri retrivial from the contextual information
4. no extends, no self, no contextual information => fail, there's no way to locate the OAI document
5. extends, self, no contextual information => do uri retrievial with extends, if it's a relative uri, use self to build the uri
6. extends, self, contextual information => do uri retrieval with extends, then try the contextual information (tool specific), if the uri is relative, use self like in 5 to build the uri
7. no extends, self, contextual information => do url retrieval from context information, if the uri is relative, use self like in 5 to build the uri
8. no extends, self, no contextual information => fail, there's no way to locate the OAI document

In any of those cases where uri retrivial is performed, there's a security risk.
In any of those case where self is present, the tool should *additionally* perform the comparison with the self of the OAD when present. Either with the full URI from extends, or with the resolved uri from overlay self + extends/contextual relative information.

Is that a faire representation? if so, do you think this is captured well enough in the current proposal?

Lastly, cases 1, 2, 3, 4 were already specified by earlier version of the specification. Cases 5, 6, 7, 8 are being added here.

versions/1.2.0-dev.md
```diff
@@ -69,7 +69,7 @@ Overlay Description authors SHOULD consider how text using such extensions will
 
 Implementations MUST NOT treat an `extends` value as unresolvable before checking all possible target [[OpenAPI]] Description documents provided to the implementation.
 
-To ensure interoperability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value MUST match the target's `$self` URI.
+To ensure portability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value SHOULD match the target's `$self` URI.
```
handrews: [I also apologize for failing to hit send on this two days ago]

I apologize for not noticing this earlier, but I'm not sure "interoperability" is the ideal word here.

This is really more about portability (of the documents) than interoperability.  Given the same input, implementations would be expected to resolve the references the same way.  I suppose this is a bit more interoperable as implementations might have different ways to figure out the current document's location for resolving a relative `extends`, but... hmm... I think this is more about portability of document sets.  idk, does anyone else think "interoperability" is better here?

versions/1.2.0-dev.md
handrews: I agree that you can drop the paragraph and things will be less confusing.

Separately, part of the point of `$self` is to allow you to provide all of the referenced (including by `extends`) documents to the tool up front so you don't have to take the performance/reliability hit or the security risk of performing actual retrieval.

versions/1.2.0-dev.md
baywet: portability works great! I believe we have already applied the suggestion, anything else?

versions/1.2.0-dev.md
baywet:
```suggestion
```

versions/1.2.0-dev.md
```diff
@@ -1,596 +1,785 @@
...
+While the framing of CommonMark 0.27 as a minimum requirement means that tooling MAY choose to implement extensions on top of it, note that any such extensions are by definition implementation-defined and will not be interoperable.
+Overlay Description authors SHOULD consider how text using such extensions will be rendered by tools that offer only the minimum support.
+
+### Parsing Documents
+
+Implementations MUST NOT treat an `extends` value as unresolvable before checking all possible target [[OpenAPI]] Description documents provided to the implementation.
+
+To ensure portability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value SHOULD match the target's `$self` URI.
+Implementations MUST examine all available target documents for matching `$self` values before concluding that no document matches the `extends` URI.
+
```
baywet:
```suggestion
```

handrews (commented):
GitHub is acting weird- it still shows my "sorry I didn't hit send two days ago" as not having been sent despite it obviously having been.

More importantly, it's showing the entire Markdown file as deleted and re-added?

I'm pretty sure tha changes are correct now based on the discussion, but the diff is unusable and I'm worried something went wrong with a recent commit?

baywet:
> More importantly, it's showing the entire Markdown file as deleted and re-added?

It looks like the commit I had made from the UI messed up the line returns somehow. I reapplied that commit correctly and force-pushed. The diff looks ok now. Thanks for calling it out!

handrews (approved):
SHIP IT! 😁

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,38 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
 While the framing of CommonMark 0.27 as a minimum requirement means that tooling MAY choose to implement extensions on top of it, note that any such extensions are by definition implementation-defined and will not be interoperable.
 Overlay Description authors SHOULD consider how text using such extensions will be rendered by tools that offer only the minimum support.
 
+### Parsing Documents
+
+Implementations MUST NOT treat an `extends` value as unresolvable before checking all possible target [[OpenAPI]] Description documents provided to the implementation.
```
ralfhandl: Can we make this a positive statement, like

```suggestion
Implementations MUST check all possible target [[OpenAPI]] Description documents provided to the implementation before treating an `extends` value as unresolvable.
```

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,38 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
+To ensure portability, when a target [[OpenAPI]] Description defines `$self`, the Overlay's `extends` value SHOULD match the target's `$self` URI.
+Implementations MUST examine all available target documents for matching `$self` values before concluding that no document matches the `extends` URI.
```
ralfhandl: This sentence seems to be mostly a rephrasing of the first sentence, with the additional info to match `$self` values - can these be combined into one sentence or paragraph?

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,38 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+- If the [`$self`](#overlay-self) field is not present, the base URI MUST be determined from the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2). The most common base URI source in this case is the retrieval URI of the Overlay document (per [RFC3986 Section 5.1.3](https://tools.ietf.org/html/rfc3986#section-5.1.3)), though other sources such as encapsulating entities ([RFC3986 Section 5.1.2](https://tools.ietf.org/html/rfc3986#section-5.1.2)) or application-specific defaults ([RFC3986 Section 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.4)) MAY apply.
+
+#### Resolving URI Fragments
+
+Where a URI appears in contexts such as CommonMark hyperlinks within `description` fields and contains a fragment identifier, the fragment MUST be resolved per the fragment resolution mechanism of the referenced document.
```
ralfhandl: Why do we state this here? Is Overlay tooling supposed to resolve URI fragments in CommonMark hyperlinks?

I'd rather remove this sentence.

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,38 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
+For Overlay Object fields that identify whole documents (`$self` and `extends`), fragment identifiers MUST NOT be used.
+
+#### Relative URI References in CommonMark Fields
+
+Relative references in CommonMark hyperlinks (such as those in `description` fields) are resolved in their rendered context, which might differ from the context of the Overlay document.
```
ralfhandl, line 98: Do we expect Overlay documents to be rendered?

We know that OpenAPI documents are often rendered as API documentation, so the sentence makes sense there, from where it was copied.

ralfhandl (commented):
Some suggestions

versions/1.2.0-dev.md
handrews: @ralfhandl I probably agree with you now that you mention it.  This is only relevant if Overlays use CommonMark _in a way that a tool is expected to render._  If Overlays only use them in the sense of changing OAS CommonMark fields, then this is not necessary (Overlays should consider the content a black box, and just render the raw string in visual editors, etc.)

baywet: One thing that wasn't clear for me was for all the description fields (in overlay objects). Do we expect them to be rendered at any point? If not, this is moot, but [because we have a common mark section](https://spec.openapis.org/overlay/v1.1.0.html#rich-text-formatting) I thought it was important to port that part as well.

versions/1.2.0-dev.md
baywet: provided a reply here https://github.com/OAI/Overlay-Specification/pull/322#discussion_r3225917548

versions/1.2.0-dev.md
baywet: You mean lines 70 and 73, correct?

baywet:
```suggestion
```

(to be consistent with the other rewording)

versions/1.2.0-dev.md
handrews: > Where Overlay tooling renders rich text

looks like it does some rendering so, yes, this probably should be included.  I'm not sure _why_ overlay tooling would be doing this rendering, but clearly someone expects it to.

baywet: Thanks for confirming the impression I had. I'll resolve this thread for now.

handrews (approved):

mikekistler (approved):
Looks good. 👍

versions/1.2.0-dev.md
```diff
@@ -65,9 +65,37 @@ Where Overlay tooling renders rich text it MUST support, at a minimum, markdown
...
+
+- If the [`$self`](#overlay-self) field is present and is an absolute URI, the base URI is the `$self` URI (per [RFC3986 Section 5.1.1](https://tools.ietf.org/html/rfc3986#section-5.1.1): Base URI Embedded in Content).
+- If the [`$self`](#overlay-self) field is present and is a relative URI-reference, it MUST first be resolved against the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2) before being used as the base URI for resolving other relative references.
+- If the [`$self`](#overlay-self) field is not present, the base URI MUST be determined from the next possible base URI source per [RFC3986 Section 5.1.2 - 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.2). The most common base URI source in this case is the retrieval URI of the Overlay document (per [RFC3986 Section 5.1.3](https://tools.ietf.org/html/rfc3986#section-5.1.3)), though other sources such as encapsulating entities ([RFC3986 Section 5.1.2](https://tools.ietf.org/html/rfc3986#section-5.1.2)) or application-specific defaults ([RFC3986 Section 5.1.4](https://tools.ietf.org/html/rfc3986#section-5.1.4)) MAY apply.
```
ralfhandl: Please make the text starting with "The most common base URI" a paragraph after the bullet list and point out that it applies to both the second and third bullet.

versions/1.2.0-dev.md
ralfhandl: Please remove this text here, similar but unfortunately and confusingly not identical text is already being added to the section on CommonMark.

ralfhandl (changes requested):

versions/1.2.0-dev.md
baywet: fixed

versions/1.2.0-dev.md
baywet: fixed

ralfhandl (approved):

lornajane:
Are we waiting for anything more here? I had a look at the wording and couldn't see anything I thought needed fixing. You've got approvals from Henry and Ralf, plus a recent but technically outdated approval from @mikekistler but I think we're in good shape here, should I approve and merge?

baywet:
@lornajane I think we're good on this one, it's been reviewed thoroughly multiple times by different people. The only thing that was missing is your editorial review, and if that's also good, let's merge this! 

Thanks

lornajane (approved):
Thank you!

PR #377, opened 2026-08-08
Add Bielu.AspNetCore.ApiDescriptions to the list of Overlay tools

bielu:
## Summary
- Adds [Bielu.AspNetCore.ApiDescriptions](https://github.com/bielu/Bielu.AspNetCore.ApiDescriptions) to the "Tools that Support Overlays" list in the README.

## Test plan
- N/A (documentation-only change)


lornajane:
Thanks! I think we need to discuss this as a group since this is not an OpenAPI tool, and Overlay is specifically scoped to OpenAPI only - it is expected that format-aware features may be introduced moving forward. I'll keep this pull request open so we can get some feedback.

PR #404, opened 2026-09-22
fix: removes duplicate abstract now covered by config

baywet:
removes duplicated abstraction sections now covered by the respec abstract options, this also should fix the numbering issue for the sections https://github.com/OAI/spec.openapis.org/pull/138#discussion_r4073687427


miqui:
@baywet looks good to me....and I can't add a formal approval on this repo.

